PR #9, opened 2020-04-16
basic utility scripts for install and watch commands

mogoodrich:
Inspired by some scripts that @jmbabazi  wrote, I created some basic shell scripts for easing running mvn install and watch on the distro projects and the parent PIH-EMR project.

My shell scripting experience is limited, so please review and provide feedback. Once we get a pattern we like, we can merge and then copy this to all distro projects.

A few notes:

* I set the "delaySeconds" to 1, because while testing I discovered that a build was never triggered until IntelliJ lost focus (assumedly because IntelliJ doesn't flush to disk until it loses focus), so having only a 1 second delay wasn't too fast, and was actually exactly what I was looking for... when you switch to the browser to test it basically immediately builds

* @mseaton  if it's easy it might make sense to have a "watch" trigger an install first...otherwise when you start working in the morning if you pull in any changes before starting the watch and forget to do an install, you won't get those changes until the next change you make.

See PR for related changes to PIH-EMR config here:

https://github.com/PIH/openmrs-config-pihemr/pull/12/files

Also certainly open to better ways to handle finding "config-pihemr" from "config-zl".


install.sh
```diff
@@ -0,0 +1,17 @@
+if [ $# -eq 0 ]; then
```
brandones, line 1: You're missing a shebang. `#!/bin/bash`

install.sh
```diff
@@ -0,0 +1,17 @@
...
+    echo "Please provide the name of the server to install to as a command line argument"
+    echo "For example: './install.sh mirebalais'"
+    exit 1
+fi
+
+# if there's a "config-pihemr" or "openmrs-config-pihemr" directory at the same level as this project,
+# run the install for it
+if [[ -f '../config-pihemr/install.sh' ]]; then
+    (cd ../config-pihemr && ./install.sh)
```
brandones, line 10: Why not make this script independent of the script being in `config-pihemr`? Wrap the `mvn` command in a function like so:
```
mvn_compile () {
  mvn clean compile -DserverId=$1
}
```

And then you can do e.g.
```suggestion
    (cd ../config-pihemr && mvn_compile $1)
```

Then these scripts will be self-contained.

watch.sh
```diff
@@ -0,0 +1,20 @@
+
```
brandones, line 1: Shebang

watch.sh
```diff
@@ -0,0 +1,20 @@
...
+# NOTE: this "watchs" the directory and rebuilds on changes, so it will run indefinitely until you "Ctrl-C"
```
brandones, line 2: Why not just include this information in the "usage" below?

watch.sh
```diff
@@ -0,0 +1,20 @@
...
+
+if [ $# -eq 0 ]; then
+    echo "Please provide the name of the server to install to as a command line argument"
+    echo "For example: './install.sh mirebalais'"
+    exit 1
+fi
```
brandones, line 8: Consider
```suggestion
usage () {
    echo "watch.sh [SERVER]"
    echo "Installs the configuration to SERVER. Watches the directory and rebuilds on changes."
    echo "Also watches parent config if it is at ../openmrs-config-pihemr or ../config-pihemr."
    echo "Ctrl+C to exit."
    echo "Example: ./watch.sh mirebalais"
}
if [ $#  -ne 1 ] || [ "$1" = "-*" ]; then
    usage
    exit 1
fi
```

watch.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# if there's a "config-pihemr" or "openmrs-config-pihemr" directory at the same level as this project,
+# run the watch for it as well
+if [[ -f '../config-pihemr/watch.sh' ]]; then
+    (cd ../config-pihemr && ./watch.sh &)
```
brandones, line 13: As above with `./install.sh`, just factor out into a function.

install.sh
mogoodrich: Oh, right!

install.sh
mogoodrich: There's a plus/minus to both approaches... it's self-contained, but now if you want to change the install instructions for the PIH EMR config you need to change it both in that install.sh and here... but I do agree, that reads cleaner, will take a look again tomorrow.

mogoodrich: Also could make sense just to remove the install.sh and watch.sh from PIH EMR because they aren't really needed (maybe that was part of what you were suggesting, actually)

mogoodrich:
Thanks for the feedback @brandones I basically made all of your changes.

Will copy this over to the other child repos after I use it bit and see if there's anything I missed.

PR #10, opened 2020-04-22
UHM-4445: Migrate Haiti, Liberia, and Sierra Leone address bundles to…

mogoodrich:
… their respective config files


brandones:
Where did this come from? I can't find a Haiti/Mirebalais/ZL address bundle in pihcore.

brandones:
Oh! Nearly a year and a half on MI/SIS and I hadn't heard of [openmrs-modules-haiticore](https://github.com/openmrs/openmrs-module-haiticore)

mogoodrich:
Yeah @brandones  Haiti Core was where we moved some of the Haiti configuration back when we were working with the Esante Plus developers and wanted a common place to share metadata/configuration  (Esante Plus is another OpenMRS instance used in Haiti, and potentially the national system).  We haven't talked to them in quite some time, and it's unclear if they are still using the module... but, long story short, as we phase out Haiti Core in our implementation at some point we may want to have a conversation with them as to if we want to share metadata in another way going forward.  This is also why I didn't remove the hierarchy information from Haiti Core.

PR #11, opened 2020-05-07
UHM-4439: Migrate GlobalPropertyBundle out of code and into config (WIP)

mogoodrich:


configuration/globalproperties/gp_zl.xml
```diff
@@ -1,8 +1,9 @@
 <config>
+    <!-- TODO: this can just be moved into the main gp.xml in PIH EMR -->
```
brandones, line 2: You have an open PR (https://github.com/PIH/openmrs-config-pihemr/pull/33) for the main gp.xml, why don't you just move it over?

mogoodrich: This is only a draft (I couldn't find the draft setting before, but I just saw it now... might have to step away from this a bit to work on mUzima so this is basically just a commit of exactly where I was at at the moment).

PR #13, opened 2020-05-27
simple utility script to pull in both PIH EMR repo and latest repo

mogoodrich:
Just a simple script that does a git pull on both the config-zl repo and the config-pihemr repo if it's in a directory relative to the ZL directory.

So the daily "update my environment" routine would be:

cd [config-zl directory]
mvn openmrs-sdk:pull
mvn openmrs-sdk:deploy
./pull.sh
./install.sh [serverId]
mvn openmrs-sdk:run

Would be great to get a another layer on top of this at some point, but this seemed like a quick-win... I can copy it to the other config directories as well.

### DIFF
--- a/pull.sh
+++ b/pull.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# if there's a "config-pihemr" or "openmrs-config-pihemr" directory at the same level as this project,
+# run a pull there as well
+
+if [[ -d '../config-pihemr' ]]; then
+    (cd ../config-pihemr && git pull)
+elif [[ -d '../openmrs-config-pihemr' ]]; then
+    (cd ../openmrs-config-pihemr && git pull)
+else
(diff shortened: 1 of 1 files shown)


lnball:
I do this manually.  This looks good.

brandones:
https://github.com/PIH/pih-emr-invoke
```
~/C/pih (master|✔) $ inv run --help
Usage: inv[oke] [--core-opts] run [--options] [other tasks here ...]

Docstring:
  Pulls, deploys, enables modules, and then runs OpenMRS.
  Accepts default answers for openmrs-sdk:deploy.

  The `--env` flag can be used to change the active server and
  run that server at the same time. To just run a particular server,
  leaving the active server as-is, you can use the `--server` flag.

Options:
  -e STRING, --env=STRING
  -i, --skip-enable-modules
  -k, --skip-deploy
  -o, --offline
  -s, --skip-pull
```

PR #15, opened 2020-08-04
UHM-4683 Move Location Tag Setup out of PIH Core and into Config

brandones:
https://pihemr.atlassian.net/browse/UHM-4683

This one's a doozie. I really think we should migrate to Iniz for all tags except `Login Location` as soon as we migrate locations.


lnball:
@brandones It should be "Centre de Santé de Lacolline de Lascahobas".  It's not a hospital.  The map on this page shows the names and locations:  https://www.pih.org/country/haiti  Unfortunately you can't copy/paste and the UPPERCASE makes it difficult to determine accents.  The "Clinique Sante Famn" is down the road in "downtown" Lascahobas.  So it's a different facility.

brandones:
Unfortunately this isn't a very good moment at which to change facility names around. Those are the names that are in the PihCore metadata: https://github.com/PIH/openmrs-module-pihcore/blob/cc48b8778062365d98896fdf79ec95f90e4b21db/api/src/main/java/org/openmrs/module/pihcore/metadata/haiti/PihHaitiLocations.java#L54 . We can change those as a separate effort. It will be easier once we implement https://pihemr.atlassian.net/browse/UHM-4354 (Move Locations etc into Iniz).

mogoodrich:
@brandones the only reason I created the ability to do locations via config is because of the need to change tags depending on site, so I'm fine with going to Iniz for most of the tags if that's easier... that being said, I feel like it's not much different, especially if we could properly set up a hierarchy for HIV and Mental Health to inherit from Haiti, because I think those are all the same (or at least could be the same)

mogoodrich:
Yeah @lnball  as @brandones  says that's the name we have for Lacolline for some reason... we should change it if it's incorrect.

But yeah, @brandones this would a good reason to change the tags to be installed via Initializer, as changing the name now would break the tags....

brandones:
@mogoodrich What's a good subset of the environments for which to diff the tag mappings? humci and what else?

mogoodrich:
@brandones you mean staging servers to review?  I'd just review the ones that are auto-deployed from Bamboo:

ami - Haiti Mental Health
CES - CI - Mexico
ci - Haiti general site
gladi - Sierra Leone
HUM-CI - Mirebalais
Kouka - Liberia
Lespwa - Haiti HIV
Peru - Peru 

May not be necessary to review *all* of them, but I think covers all our basic configurations?

brandones:
Okay. I'm waiting on access to ami, ci, and lespwa.

brandones:
Okay, I've made exports of location_tag_map for ami, ci, humci, and lespwa.

PR #17, opened 2020-08-08
Restore tags to exactly what they were on the servers before

brandones:
https://pihemr.atlassian.net/browse/UHM-4683


mogoodrich:
@brandones  I am going to go ahead and merge this in because @lnball  is blocked since it looks like the HIV forms are not available.

PR #21, opened 2020-09-10
UH-3686: enrollInProgram should set the location

mogoodrich:


mogoodrich:
fyi @lnball  (though note that this won't work until the changes I made to HFE get merged in)

PR #26, opened 2020-09-28
UHM-4851, display last prescribed regimen

cioan:
Hi @mogoodrich , I saw you have used REST and AJAX calls in other html forms, so I follow that example to add a small REST call to get the last prescribed regimen obsgroup. Thanks for reviewing!


mogoodrich (approved):
Thanks @cioan ... I had forgotten about using REST calls within a form, but, yeah, that seems like a great approach for this situation... LGTM!

cioan:
Thanks @mogoodrich !

PR #29, opened 2020-10-08
UHM-4833: Don't allow adding encounters to visits at a different loca…

mogoodrich:
…tion from the visit location (?)

Set Unknown location as an HIV Consult, Lab Results and Vitals Location


mogoodrich (commented):
We are restricting encounter locations so that they must match visit location.  

As we have several encounter and visits we are migrating from the old HIV system with an "Unknown Location" we must make sure for all the encounter types that we are migrating over "Unknown Location" is allowed... otherwise you would not be able to edit one of these migrated encounters without getting a validation error, nor would you be able to add a new encounter to an existing Visit at an Unknown Location.

lnball (commented):
Looks reasonable, but will let Mike weigh in.

mseaton (approved):

PR #30, opened 2020-10-09
Changes to key population answers

jmbabazi:


mogoodrich (approved):
i didn't review all the concepts in detail, but generally LGTM!

jmbabazi:
Thanks @mogoodrich, I believe there is no hurry here. We can wait to hear from @lnball.

lnball (changes requested):
@jmbabazi Wrap these in a repeat/template.  Easier for readability and updates.  The only changese are the mode/cId.  You can use the previous render lines, but change the conceptId, replace the answers with u/n/u, and format in a table/rows/cells.

jmbabazi:
@lnball , be

PR #32, opened 2020-10-16
UHM-4935, use the encounterDate datepicker widget value

cioan:
@mogoodrich , here is, I think a safer way to get the encounter date of the form. I am not sure what changed, I am pretty sure the encounter lookup expression used to work for new encounters, either in the past or present. But, anyway this PR addresses this problem. Thanks.

### DIFF
--- a/configuration/pih/htmlforms/hiv/zl/hiv-dispensing.xml
+++ b/configuration/pih/htmlforms/hiv/zl/hiv-dispensing.xml
@@ -104,7 +104,7 @@
 
 <script type="text/javascript">
 
-    jq(function() {
+    jq(document).ready(function() {
 
         var artTreatmentConstructUuid = '0b1ca3d0-9f70-4366-a58c-c0f7491c47da';
         var medicationOrdersUuid = '3cd9491e-26fe-102b-80cb-0017a47871b2';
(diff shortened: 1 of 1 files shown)


configuration/pih/htmlforms/hiv/zl/hiv-dispensing.xml
```diff
@@ -120,10 +120,21 @@
 
         var currentEncounterDate = new Date();
         var encounterDate = '<lookup expression="encounter.getEncounterDatetime().getTime()"/>';
-        if (typeof encounterDate !== "undefined" &amp;&amp; encounterDate !== null &amp;&amp; (encounterDate.length > 0)) {
-          currentEncounterDate = new Date(+encounterDate);
+          if (typeof encounterDate !== "undefined" &amp;&amp; encounterDate !== null &amp;&amp; (encounterDate.length > 0)) {
+            currentEncounterDate = new Date(+encounterDate);
+        } else {
+          // look for the encounterDate datepicker widget
+          var encounterDateValue = jq("#encounterDate .hasDatepicker");
```
mogoodrich, line 127: Does the getField method not work here?  (Not a big deal if it doesn't, just curious)

configuration/pih/htmlforms/hiv/zl/hiv-dispensing.xml
```diff
@@ -185,11 +196,7 @@
 
 
         var validateNextDispensingDate = function() {
-            var currentEncounterDate = new Date();
-            var encounterDate = '<lookup expression="encounter.getEncounterDatetime().getTime()"/>';
-            if (typeof encounterDate !== "undefined" &amp;&amp; encounterDate !== null &amp;&amp; (encounterDate.length > 0)) {
-                currentEncounterDate = new Date(+encounterDate);
-            }
+            var currentEncounterDate = new Date(getField('encounterDate.value').datepicker('getDate'));
```
mogoodrich, line 199: Just new Date(getField('encounterDate.value'))" isn't enough?

Again, just curious, not a big deal if it works as-is...

mogoodrich (approved):

configuration/pih/htmlforms/hiv/zl/hiv-dispensing.xml
cioan: Thanks @mogoodrich ! Yes, I have noticed in the javascript console that sometimes when the form loads it throws a null getField() method not found. I have tried to debug it, but when I set a breakpoint it works b/c there is plenty time for the page and all depending js scripts to load. When I switched to using jquery selector I did not see that error anymore.

mogoodrich: Interesting... something to keep an eye on for other places where it may cause issues.

PR #33, opened 2020-10-26
UHM-4922: Create data warnings ETL/DW table

mogoodrich:


configuration/pih/petl/jobs/datawarnings/aggregrate/job.yml
```diff
@@ -0,0 +1,16 @@
...
+    datasource: "mysql/openmrs.yml"
+    query:  "datawarnings/aggregrate/source.sql"
+
+  load:
+    datasource: "sqlserver/openmrs_extractions.yml"
+    table: "hivmigration_aggregrate_data_warnings"
+    schema: "datawarnings/aggregrate/target.sql"
+
+  dropAndRecreateTable: false
+  conditional: "SELECT COUNT(*)=1 FROM information_schema.tables WHERE table_schema = 'openmrs' AND table_name = 'hivmigration_data_warnings'"
```
mogoodrich, line 13: Note that this depends on the database being names "openmrs"... which should be fine for all our production instances, but may cause issues for our local dev environments.  I was thinking we should support some sort of environ variable search and replace with the conditional (and all the scripts, for that matter), but didn't add it at this point.

configuration/pih/petl/jobs/datawarnings/aggregrate/source.sql
```diff
@@ -0,0 +1,6 @@
+SELECT
+    warning_type,
+    count(*) as count,
+    NOW() as timestamp
+FROM hivmigration_data_warnings
+GROUP BY warning_type
```
mogoodrich, line 6: This is pretty simple pattern, but it's possible that it could be used for other simple aggregate queries:  note that it uses a group by and also added a timestamp column. "dropAndRecreateTable" has been set to false, so on each run the it inserts a row per unique warning into the target table with the counts of each unique.  So, as data errors are resolved, this should be a way to track the decrease in errors over time.

configuration/pih/petl/jobs/datawarnings/aggregrate/job.yml
mogoodrich: (Also the use case for this is for the other, non-HIV migration servers in the ZL network... we don't want this job to fail just because that table isn't present on those servers)

cioan (approved):
Thanks @mogoodrich !

configuration/pih/petl/jobs/datawarnings/job.yml
```diff
@@ -0,0 +1,15 @@
...
+  extract:
+    datasource: "mysql/openmrs.yml"
+    query:  "datawarnings/source.sql"
+
+  load:
+    datasource: "sqlserver/openmrs_extractions.yml"
+    table: "hivmigration_data_warnings"
+    schema: "datawarnings/target.sql"
+
+  conditional: "SELECT COUNT(*)=1 FROM information_schema.tables WHERE table_schema = 'openmrs' AND table_name = 'hivmigration_data_warnings'"
```
mseaton, line 12: If the 95% use case for the conditional is going to be to check the existence of a table, would be helpful to have a more convenient way to express that, though this is fine for now.  That could even just be creating a "table_exists" function that takes in the table name and returns a boolean, which would allow the conditional to be "select table_exists('hivmigration_data_warnings')"

mseaton (approved):
LGTM, just one thought comment.

configuration/pih/petl/jobs/datawarnings/job.yml
mogoodrich: Makes sense, thanks @mseaton

PR #34, opened 2020-10-29
UHM-4922: Create data warnings ETL/DW table

mogoodrich:


configuration/pih/petl/jobs/datawarnings/aggregrate/job.yml
```diff
@@ -9,7 +9,7 @@ configuration:
     table: "hivmigration_aggregrate_data_warnings"
     schema: "datawarnings/aggregrate/target.sql"
 
-  dropAndRecreateTable: false
+  dropAndRecreateTable: true
```
mogoodrich, line 12: We need to drop and recreate the table because of the schema change... will switch this back to false after the first build.

cioan (approved):

PR #37, opened 2020-11-18
Sort component names

brandones:


mogoodrich (approved):
+1 

Thanks for doing it... I've been meaning to do something like this for a while, as it will allow easier comparisions between what is installed on the servers.

I didn't actually review that the sorting was correct and that nothing was inadvertently dropped/added... I assume you're reasonably confident you got it right? :)

lnball:
Thanks for cleaning up after us @brandones

PR #39, opened 2020-11-24
Add Consult Location tag to all locations in pih-config-haiti-hiv

brandones:
Part of https://pihemr.atlassian.net/browse/UHM-4979


mogoodrich (approved):
LGTM... it's probably fine to just use "HIV Consult Location" for this form (but there's nothing really wrong with adding the Consult Note Location tag as well).

PR #40, opened 2021-01-08
Initial pass at demonstrating drug order tag usage for ART.

mseaton:
Just creating this PR to demonstrate what I have in place so far.  I'm not ready to commit yet, as there are a few issues still to work out.  But figured I could use this PR to show what the tag config looks like so far, and throw a few screenshots up for review.  Please let me know if you have any feedback.

image

image

image

image

### DIFF
--- a/configuration/pih/htmlforms/hiv/section-hiv-plan.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-plan.xml
@@ -847,10 +847,10 @@
                         <option value="0fc3d5c1-fd39-4899-b5c0-b094e28ff359" label="DRV/ETV 300/100 mg"/>
                     </orderProperty>
                     <orderProperty name="action" label="Action">
-                        <option value="NEW" label="Place a new Order"/>
-                        <option value="REVISE" label="Revise this Order"/>
-                        <option value="RENEW" label="Renew this Order"/>
-                        <option value="DISCONTINUE" label="Discontinue this Order"/>
+                        <option value="NEW" label="New Order"/>
+                        <option value="REVISE" label="Revise Order"/>
+                        <option value="RENEW" label="Renew Order"/>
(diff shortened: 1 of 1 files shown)


lnball:
This looks good @mseaton .  Is the first screenshot the view?  Or this is on the edit page?  I know the view wasn't working yesterday, so it's fine if that's not completed yet.

mseaton:
@lnball yes the first screenshot is the view.  There was a tweak needed to pihcore which I added yesterday.

mogoodrich:
Generally looks great to me!

A few functionality questions, but this is really me guessing on how we think we'd want it to be used, so definitely more a starting point for discussion rather than anything else

* We probably want to hide the "NEW" in the display (which I think is the order type), especially if we go with open-ended orders. ie, if a drug was started in 2018 as an open ended form, if a new follow up form was opened today, we wouldn't want it to say "NEW", just:

AZT/3TC/NVP

Started on January 7, 2018
etc...

* If we do go with the open-ended order approach, can we skip the quantity/refills/etc, as this seems less relevant in those cases? (Apologies if we discussed this yesterday)


* That all being said, it is likely better/more accurate to not use the open-ended approach.  If so, does the tag provide a "renew" option?  (Apologies again if we discussed this yesterday).  This seems like the correct way to handle orders... during a follow-up the doctor is shown the current orders for a patient and can 1) do nothing if the order is still active and no changes need to be made to the drug, 2) renew if the order has expired/patient needs refills but no changes need to be made, 3) revise if a tweak needs to be made to the drug or dosage, etc.  But I'm thinking this through from a POC approach and how things work when I visit my doctor for a follow-up.  I think it'd probably best, if possible, to design for POC as that's our long-term strategy, but we'd need to make in the short-term that data clerks have the information on paper to fill out this data retrospectively.

@mseaton  @dmdesimone  @lnball 

(Not sure if it's best to move this discussion to a ticket or something other than a PR)

PR #41, opened 2021-01-08
Change concept used for "TB continue treatment", see UHM-5103

mogoodrich:
See discussion in comments section here:
https://pihemr.atlassian.net/browse/UHM-5103


lnball (approved):
Looks good

PR #42, opened 2021-01-13
UHM-3348, update Haiti locations

cioan:
Did I miss any other config files?


brandones (approved):
LGTM! Regarding your question, I trust the global find-replace :)

mogoodrich:
Just getting to this now, but looks good... and agree re: global search and replace... the other good thing is, at least for HUM-CI, CI and Ami, if there's a tag issue the smoke tests would probably have failed hard (they didn't).

We should just try to remember to look out for this when we next update our Haiti servers... fyi @jmbabazi ... the other good thing is if there is a mistake, since the tags are in the config it will be easier to "hot" fix it...

PR #43, opened 2021-01-13
UHM-5093 Capture HIV Contacts for new HIV patients

brandones:
Please see also https://github.com/PIH/openmrs-config-pihemr/pull/109

https://pihemr.atlassian.net/browse/UHM-5093

Screenshot from 2021-01-13 15-03-25
Screenshot from 2021-01-13 15-05-14

### DIFF
--- a/configuration/pih/htmlforms/hiv/section-hiv-history.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-history.xml
@@ -754,17 +754,22 @@
                     <div>
                         <obs conceptId="PIH:AGE" labelCode="pihcore.age"/>
                     </div>
-                    <div>
-                        <obs conceptId="CIEL:163533"
-                             labelCode="pihcore.dead"
-                             answerConceptIds="PIH:ALIVE,PIH:DEATH,PIH:UNKNOWN"
-                             style="checkbox"/>
+                    <div style="margin-right: 64px;">
+                        <label><uimessage code="pihcore.dead"/></label>
+                        <p class="radio">
+                            <obs conceptId="CIEL:163533"
+                                answerConceptIds="PIH:ALIVE,PIH:DEATH,PIH:UNKNOWN"
+                                answerCodes="general.no,general.yes,pihcore.unknown"
+                                style="radio"/>
+                        </p>
                     </div>
                     <div>
-                        <obs conceptId="PIH:RESULT OF HIV TEST"
-                             labelCode="pihcore.hiv.serologicalStatus"
-                             answerConceptIds="PIH:UNKNOWN,PIH:POSITIVE,PIH:NEGATIVE"
-                             style="checkbox"/>
+                        <label><uimessage code="pihcore.hiv.serologicalStatus"/></label>
+                        <p class="radio">
+                            <obs conceptId="PIH:RESULT OF HIV TEST"
+                                answerConceptIds="PIH:UNKNOWN,PIH:POSITIVE,PIH:NEGATIVE"
+                                style="radio"/>
+                        </p>
                     </div>
                 </obsgroup>
             </div>
--- a/configuration/pih/htmlforms/hiv/section-hiv-history.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-history.xml
@@ -741,9 +741,11 @@
             <div class="section-container mb-4 flex-questions">
                 <obsgroup groupingConceptId="PIH:Contact construct">
                     <div>
-                        <obs conceptId="PIH:RELATIONSHIP OF RELATIVE TO PATIENT"
+                        <!-- Relationship to patient -->
+                        <obs conceptId="CIEL:164352"
                          labelCode="pihcore.contactRelationship"
-                         answerConceptIds="PIH:PARTNER OR SPOUSE,PIH:GUARDIAN RELATIONSHIP TO CHILD,PIH:SIBLING,PIH:OTHER RELATIVE"/>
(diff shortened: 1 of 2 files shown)


configuration/pih/htmlforms/hiv/section-hiv-history.xml
```diff
@@ -711,6 +721,42 @@
...
+                    <div>
+                        <obs conceptId="PIH:LAST NAME" labelCode="pihcore.lastName"/>
+                    </div>
+                    <div>
+                        <obs conceptId="PIH:AGE" labelCode="pihcore.age"/>
+                    </div>
+                    <div>
+                        <obs conceptId="CIEL:163533"
+                             labelCode="pihcore.dead"
+                             answerConceptIds="PIH:ALIVE,PIH:DEATH,PIH:UNKNOWN"
```
lnball, line 746: The paper form is different, but I mostly like this.  I would suggest using answerCodes=yes/DEATH, no/ALIVE, unknown/unknown.  This keeps it similar to the paper, but adding unknown.  Otherwise it looks odd with the header or "death" with dead.   Too knit-picky?

configuration/pih/htmlforms/hiv/section-hiv-history.xml
```diff
@@ -711,6 +721,42 @@
...
+                             style="checkbox"/>
+                    </div>
+                    <div>
+                        <obs conceptId="PIH:RESULT OF HIV TEST"
+                             labelCode="pihcore.hiv.serologicalStatus"
+                             answerConceptIds="PIH:UNKNOWN,PIH:POSITIVE,PIH:NEGATIVE"
```
lnball, line 752: Not sure of the subtle difference between unknown and Indeterminate, but the paper form and concept have "Indeterminate", positive and negative.  I would modify to use Indeterminate.

lnball (changes requested):

configuration/pih/htmlforms/hiv/section-hiv-history.xml
brandones: Not at all, that's great feedback, I agree. Fixed it up. I'd also messed up the formatting—I'd intended for these to be radio buttons rather than dropdowns. I've fixed that as well.

configuration/pih/htmlforms/hiv/section-hiv-history.xml
```diff
@@ -711,6 +721,47 @@
...
+                                answerConceptIds="PIH:ALIVE,PIH:DEATH,PIH:UNKNOWN"
+                                answerCodes="general.no,general.yes,pihcore.unknown"
+                                style="radio"/>
+                        </p>
+                    </div>
+                    <div>
+                        <label><uimessage code="pihcore.hiv.serologicalStatus"/></label>
+                        <p class="radio">
+                            <obs conceptId="PIH:RESULT OF HIV TEST"
+                                answerConceptIds="PIH:UNKNOWN,PIH:POSITIVE,PIH:NEGATIVE"
```
lnball, line 756: Did you mean to update this line too?  Indetermine?

configuration/pih/htmlforms/hiv/section-hiv-history.xml
brandones: The paper form has "inc.," which is short for "inconnue," which  which is French for "unknown."

The difference between "unknown" and "indeterminate," as I understand it, is that indeterminate is a kind of reading from a test, whereas unknown could signify an indeterminate test, no test done, or whatever else.

configuration/pih/htmlforms/hiv/section-hiv-history.xml
brandones: See above

configuration/pih/htmlforms/hiv/section-hiv-history.xml
brandones: ^^ @lnball

brandones:
Actually, I just noticed that the relationships question needs some clarifying.

screenshot--2021-01-13--191639

Presumably the question should be something like "Relationship to patient" and the possible responses should be
- "Partner or Spouse"
- "Parent or Guardian"
- "Child"
- "Sibling"
- "Other relative"

From looking at the answer set on [the question concept](https://concepts.pih-emr.org/openmrs/dictionary/concept.htm?conceptId=2172), it seems like [Guardian relationship to child](https://concepts.pih-emr.org/openmrs/dictionary/concept.htm?conceptId=1826) is intended to map to "Child" here? Which is not intuitive. That would make more sense if the question was "Relationship of patient to contact," but that would be a weird way to ask that.

Also, there's no concept for "Parent or Guardian," just "Mother" and "Father." Not sure if we should make a concept for that or just have that be the only gendered option (maybe it makes sense, with PTME and all).

Let me know what you want to do @lnball .

brandones:
I suggest for Child [CIEL:1528](https://mdsbuilder.openmrs.org/openmrs/dictionary/concept.htm?conceptId=1528) which is marked in CIEL but not in PIH as equivalent to [PIH:6396](https://concepts.pih-emr.org/openmrs/dictionary/concept.htm?conceptId=6396).

And then perhaps [CIEL:160639](https://concepts.pih-emr.org/openmrs/dictionary/concept.htm?conceptId=3565) for Guardian, which is already in PIH.

lnball:
@brandones I've changed to a new concept for relationship and cleaned up the answers.  Based on the migration data, let me know if these are wrong.  I think parent and guardian should be different answers -- especially since a parent might have more effect on the child's HIV status.
image

brandones:
Okay! Thanks @lnball . Should be in good shape. Let me know.

PR #45, opened 2021-01-21
Add htmlform (with configuration for associated form) for Drug Order …

mseaton:
…Encounter to use for migrated "regimes"

The important thing here is just that this installs the form with the given encounter type.  The details of the html and schema we can evolve later, but we need the form to exist for the migration.

### DIFF
--- a/configuration/pih/htmlforms/hiv/hiv-drug-order-documentation.xml
+++ b/configuration/pih/htmlforms/hiv/hiv-drug-order-documentation.xml
@@ -1,0 +1,155 @@
+<htmlform formUuid="96482a6e-5b62-11eb-8f5a-0242ac110002"
+          formEncounterType="0b242b71-5b60-11eb-8f5a-0242ac110002"
+          formName="HIV Drug Order Documentation" formVersion="1.0">
+
+    <style type="text/css">
+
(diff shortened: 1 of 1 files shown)


configuration/pih/htmlforms/hiv/hiv-drug-order-documentation.xml
```diff
@@ -0,0 +1,155 @@
...
+        }
+    </style>
+
+    <!-- Stolen from the dispensing module, but using prescription construct instead of dispensing -->
+    <section id="id-treatment" sectionTag="fieldset" headerTag="legend" headerStyle="title"
+             headerCode="pihcore.visitNote.orders.medications">
+
+        <div class="section-container-color">
+
+            <drugOrder format="select" selectLabel="Add another drug...">
```
lnball, line 95: All the labels should use message property codes.  Is that possible with selectLabel and label?  Or is there a selectCode or labelCode?

mseaton: Yes, you can just use a message code instead of a hard-coded label here.

mogoodrich:
LGTM!

PR #46, opened 2021-01-28
UHM-5112: For Intake and Followup Forms, migrate Form Version

mogoodrich:


brandones (approved):
I assume this is all the usual boilerplate

mseaton (approved):
I didn't really check in detail - this will come out in testing I think.  This makes me think we might want to consider some sort of HFE tag which generates it's own schema and layout based on what's in an existing encounter (and which would only work obviously in view and edit modes, not enter mode).  And it would just render Obs tags using some conventions.  Thoughts @mogoodrich ?

mogoodrich:
@mseaton  potentially, though this current PR isn't about that really.  All I did was copy the top-level encounter form for the existing HIV Intake and Followup forms and changed the version number and the uuid.  This form is just about editing the top level location/provider/date, etc... it provides no access to the underlying obs.  Really, thinking about it, these forms could just be entirely empty except for the header, because we currently don't allow any way to view/edit these old versions.  If we did then, yeah, your idea might make sense.  Will be good to get this in place so that everyone can take a look at what we currently have and how we might want to move forward.

mseaton:
Just having the forms isn't enough.  We need to allow viewing the encounters (and ideally editing them).

mogoodrich:
@mseaton  that's a bigger discussion for the whole team... :)

mogoodrich:
But I'm hoping getting this in will make it clear to everybody what is/isn't possible without actually making the old versions of the forms.

PR #47, opened 2021-02-09
UHM-5209: Handle versioning of HIV Intake and Followup sections

mogoodrich:


mogoodrich:
Here's my changes to support different versions of the "section-plan".

Overall docs of how this all works found here:
https://wiki.openmrs.org/display/docs/HTML+Form+Entry+UI+Module

But, in summary, I:
* Copied the current "section-plan.xml" into a new "section-plan_v3.0.xml" and set the version to "3"
* Changed the version of the current "section-plan.xml" to "4" and gave it a new uuid.

So now, whenever an old, migrated Follow Up encounter created with "v3" of the HIV EMR form is opened, it should use "section-plan_v3.0.xml".

For *new* follow-up forms, the default "section-plan.xml" should be used.

Right now the two forms are identical, but as changes are made one or the other we should see the different.

(@lnball  I started to try to un-comment the old stuff for use on the v3 version of the form, but then figured it was probably best to leave to you so I didn't screw anything up)

configuration/pih/htmlforms/hiv/section-hiv-plan.xml
```diff
@@ -1,6 +1,6 @@
 <htmlform id="hiv-plan"
-          formUuid="e86e3118-731c-11e7-8cf7-a6006ad3dba0" formName="HIV Plan"
```
lnball, line 2: One annoyance is that message properties for these form(lets) use this uuid.  So when you change the uuid for different versions, you also need to add a new message to message properties (and translate).  It will use the formName, but unless you add to message properties, it won't provide the translations.  As I am reviewing this file, I will add the uuids to the message property file.

ui.i18n.Form.name.e86e3118-731c-11e7-8cf7-a6006ad3dba0=HIV Plan

lnball (approved):
Otherwise...lgtm

configuration/pih/htmlforms/hiv/section-hiv-plan.xml
lnball: And sorry for the long delay...

PR #51, opened 2021-02-13
UHM-4939 - Initial pass at find patient column configuration customiz…

mseaton:
…ations for the Haiti HIV server(s)


mogoodrich (approved):
LGTM, thanks @mseaton !

PR #53, opened 2021-02-24
UHM-5106: Fix migration of Ordered ARV

mogoodrich:


configuration/pih/htmlforms/hiv/section-hiv-plan_v3.0.xml
```diff
@@ -671,83 +671,19 @@
...
-                                NVP 50 mg,
-                                TDF/3TC/EFV 300/300/600 mg,
-                                3TC 10 mg sp,
-                                ABC/3TC 60/30 mg,
-                                RAL/DRV/r 400/600/100mg,
-                                DRV/ETV 300/100 mg,
-                                ABC/3TC 120/60 mg" />
-
-                        <label>ARV #2</label>
-                        <obs id="arv2Drug" conceptId="PIH:13162"
```
mogoodrich, line 716: We are not migrating any data into this field, and now this form is only being used for old, migrated data, so I think we can remove this.

configuration/pih/htmlforms/hiv/section-hiv-plan_v3.0.xml
```diff
@@ -671,83 +671,19 @@
...
-                                ABC 60 mg,
-                                LPV/r 100/25 mg,
-                                LPV/r 80/20 mg,
-                                ABC 20 mg,
-                                LPV/r 200/50 mg,
-                                ATV/r 300/100 mg,
-                                EFV 200 mg,
-                                ETV 100mg,
-                                DRV/ETV 300/100 mg" />
+                             answerConceptIds="3cd43a64-26fe-102b-80cb-0017a47871b2,
```
mogoodrich, line 676: This form is now only being used for old, migrated data, and the only data we are migrating in are the concepts listed below.

brandones (approved):
Obviously preferable to avoid UUIDs when possible, but I imagine you did it that way for a reason. LGTM

mogoodrich:
> Obviously preferable to avoid UUIDs when possible, but I imagine you did it that way for a reason. LGTM

It's a fair point @brandones ... I chose UUIDs for two reasons:
* the existing form used UUIDS
* I was able to quickly get the UUIDs from the DB

That being said, you are right, I should switch to mappings, I will do this...

mogoodrich:
Updated to use CIEL codes instead of uuids... thanks @brandones , fyi @lnball

lnball (approved):
I'm a bit dizzy with the rationale with going back to concepts, but this code seems right.  I trust you and that you are deep into these decisions.

And yes, it's best to use CIEL mappings when possible -- and not uuids.  (Drugs don't have mappings which is the reason for using uuids.)

mogoodrich:
Thanks @lnball !

I think the reason we needed to switch back to concepts is because we only have concept-level information on the old forms... @brandones  I think you worked on functionality before I was closely involved... was there ever talk of trying to figure out how to map to action drugs without  having the dosage information...

@dmdesimone  we should also consider how the downstream pipelines work....

brandones:
No, @mogoodrich , there was no such talk. Why would we do that? We would almost certainly produce erroneous dosage information.

mogoodrich:
Yeah @brandones  I wasn't sure either how that could have worked... thanks!

PR #54, opened 2021-02-24
Add field for TB Other

brandones:
This also introduces a better way of doing a two-column layout. Unfortunately it requires a little bit of adaptation—it requires that each column be wrapped in its own div. But the result is columns that are 100% reliable, i.e. they won't shift around based on screen size, zoom, browser version, or any of that.

I just want to run this by you, @lnball , before I go ahead and change them all, at least for this form section.

The new concept is [Anti-TB (free text)](https://concepts.pih-emr.org/openmrs/dictionary/concept.htm?conceptId=13328) and was added to Haiti_HIV-161 in https://github.com/PIH/openmrs-config-pihemr/commit/57221a563db2f924f08c4952d78ca787e22b8122 (which has a completely erroneous comment, my bad).

Screenshot from 2021-02-24 13-05-16

Screenshot from 2021-02-24 13-25-17

### DIFF
--- a/configuration/pih/htmlforms/hiv/section-hiv-state.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-state.xml
@@ -10,10 +10,21 @@
                 vertical-align: top;
             }
 
-            .two-columns {
+            .two-columns-legacy {
                 column-count: 2;
                 -webkit-column-count: 2;
                 -moz-column-count: 2;
+            }
+
+            .two-columns {
+                display: table;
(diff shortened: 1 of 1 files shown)


lnball (approved):
Looks good.

NOTE:  It's not a problem (at least not for now), but the current anti-TB treatment section uses "TB drug treatment start date (CIEL:1113)" and it's not wrapped in a convset.  This isn't used for planned TB order/treatment so it's not an issue.

PR #58, opened 2021-03-01
change plan clinicial impressions and action plan to textareas

mogoodrich:
@lnball I figure these should be textareas?  (It became relevant recently because we are migrating in the "action plan" and the incoming data has carriage returns which aren't displayed in the plain "text" widget)


lnball (approved):
I thought I usually use textareas, but wouldn't have surprised me if I was sloppy.

PR #59, opened 2021-03-08
UHM-5311: Add ART treatment reason and "Has the patient started or wi…

brandones:
…ll start prophylaxis (PEP)?" questions to PIH EMR plan form version 3

https://pihemr.atlassian.net/browse/UHM-5311

I have omitted the message codes in favor of plain French text since this is just for a legacy form. Let me know if you'd rather I use message codes in this case as well.

Screenshot from 2021-03-08 11-44-15

### DIFF
--- a/configuration/pih/htmlforms/hiv/section-hiv-plan_v3.0.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-plan_v3.0.xml
@@ -42,7 +42,7 @@
             -moz-column-count: 3;
         }
 
-        p.radio > * {
+        .radio > * {
             display: inline;
             float: none !important;
             min-width: initial;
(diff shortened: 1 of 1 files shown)


lnball (changes requested):
@brandones I would recommend switching to this CIEL concept which is already in PIH EMR:  https://concepts.pih-emr.org/openmrs/dictionary/concept.htm?conceptId=10964.  I changed the datatype from boolean to coded, but it's really the same as the one you newly created.  Can you add the message prop?  It's a quick detail.

brandones:
@mogoodrich I'm trying to make the change Ellen asked for and now getting an NPE from HFE. I'm looking into it. Let me know if you need this change with urgency.

lnball:
I'll recreate the hiv mds and that should fix it.

e
________________________________

mogoodrich:
no urgency @brandones

brandones:
https://github.com/PIH/openmrs-config-pihemr/pull/140

lnball (approved):
Thanks @brandones .  Looks good.  I won't squash until you debug the NPE.

brandones:
The NPE is fixed, and this HFE code won't produce it. Thanks for the review, @lnball .

PR #64, opened 2021-03-24
UHM-5328 - Switch Prophylaxis section on v4 "Plan" to use drug orders

mseaton:
This is ready for review.  Feel free to just merge in if it'll be easier to review and test once on the test server.  This does the following:

* Replaces the prophylaxis section with a drug order tag, supporting the same orderables, minus the "other" option
* Removes the example lab order section that I had added a while back as a demonstration
* Organizes each of the order categories into sections, as it was getting hard to distinguish them - to me this looks better

We likely want to to an overhaul to the UI/UX of the drug ordering tags in general, but leaving that for a different ticket, as we will want to mock something specific up.

image

### DIFF
--- a/configuration/pih/htmlforms/hiv/section-hiv-plan.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-plan.xml
@@ -127,42 +127,6 @@
         </div>
     </section>
 
-    <style>
-        #test-orders-new .orderwidget-history-section {
-            display:inline-block;
-        }
-        #test-orders-new .order-action-section {
-            display:inline-block;
-        }
-        #test-orders-new .order-cancel-action-section {
-            display: inline-block;
-            padding-left: 10px;
-            vertical-align: middle;
-        }
-    </style>
-
-    <section id="test-orders-new" sectionTag="fieldset" headerTag="legend" headerStyle="title" headerCode="Lab Ordering Test Example">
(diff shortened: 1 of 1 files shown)


configuration/pih/htmlforms/hiv/section-hiv-plan.xml
```diff
@@ -489,224 +453,104 @@
...
-
-                        <repeat>
-                            <template>
-                                <obsgroup groupingConceptId="PIH:10742">
-                                    <tr>
-                                        <td>
-                                            <obs conceptId="CIEL:1282" answerConceptId="{conceptMap}"
-                                                answerLabel="{drugname}"
-                                                toggle="{id: '{drugname}-treat', style: 'dim'}"/>
-                                            <p class="radio {drugname}-treat">
```
lnball, line 583: This should stay.  We need to record INH treatment line (for reporting - I think).  I think the label should be changed to "INH treatment line" with primary/secondary OR "Isoniazid treatment line" with the same answers.  It won't be associated with the actual order, but I think it should be recorded in this section.

Otherwise, it should be fine @mseaton

lnball (changes requested):
Thanks for the screenshot, but I really care about the French (more than English).  Make the one change, push, and I'll test @mseaton

mseaton:
Sounds good @lnball - good catch.  The screenshot wasn't there for the language, it was there for the formatting.

PR #65, opened 2021-03-25
Uhm 5331

mseaton:
UHM-5331 - Moves radiology report questions into a new Radiology Findings section on the hiv history section (which is used by intake includes additional intake question), and into the hiv state section (which is used by followup and does not include intake question).  Modified plan and plan-v3 to collected the same 3 radiology tests as orders, using the "PIH:Radiology Image Ordered" concept.

### DIFF
--- a/configuration/pih/htmlforms/hiv/section-hiv-history.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-history.xml
@@ -1607,6 +1607,46 @@
         </div>
     </section>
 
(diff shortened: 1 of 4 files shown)


lnball (commented):
LGTM.  Since @mogoodrich had an opinion, leaving this un-approved so he can review too.

mogoodrich (approved):
LGTM, merging, and should be able to test tomorrow!

PR #68, opened 2021-03-29
Uhm 5358

mseaton:
This changes the HIV Treatment section of the clinical state to use concepts rather than drugs.  This uses the paper form as a guide, supporting each of the specific regimens listed on the paper form.  It also adds one additional regimen - TDF + 3TC + DTG - which is a more recent standard regimen that many / most patients have transitioned onto (from what I understand).   @lnball  please review from a concept perspective / @mogoodrich please let me know if anything looks off from a migration perspective.

### DIFF
--- a/configuration/pih/htmlforms/checkin.xml
+++ b/configuration/pih/htmlforms/checkin.xml
@@ -36,7 +36,7 @@
             display: inline-block;
             margin: 0px;
             margin-right: 10px;
-            color: #501d3d;
+            color: #3E3E3E;
             vertical-align: top;
             font-size:0.9em;
             }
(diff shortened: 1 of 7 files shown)


lnball:
@mseaton wrote:  "This changes the HIV **Treatment** section of the clinical state to use concepts rather than drugs."  To be exact, this is the "current treatment" section.  Happy to review now, and thanks for your work on this.

lnball (commented):
@mseaton The section-clinical-state file looks fine.  Why the other files (history, plan, planV3) with radiology changes?

mseaton:
Oops.  Must have used the wrong branch.  I'll try to fix...

mseaton:
Actually, looks like I just needed to merge up from master @lnball .  Should reflect the only change to the section-clinical-state now.

lnball (commented):
Looks good @mseaton

PR #71, opened 2021-04-02
UHM-5369 - Initial quick simple fix to reduce size of occupation list…

mseaton:
… dropdown and enable scrolling.


lnball (approved):
Seems like a quick win.  (Will ticket the different occupation list for non-HIV)  @mseaton

PR #75, opened 2021-04-13
UHM-5273, configure paper record id

cioan:


configuration/globalproperties/gp_zl.xml
```diff
@@ -48,5 +48,9 @@
             <property>labworkflowowa.labResultsEncounterTypes</property>
             <value>${encounterType.labSpecimenCollection.uuid},${encounterType.laboratoryResults.uuid},${encounterType.covidIntake.uuid},${encounterType.covidFollowup.uuid},${encounterType.zlHivIntake.uuid},${encounterType.zlHivFollowup.uuid}</value>
         </globalProperty>
+        <globalProperty>
+            <property>registrationapp.paperRecordIdentifierDefinition</property>
+            <value>paperrecord.patientDataDefinition.paperRecordIdentifier</value>
+        </globalProperty>
```
mogoodrich, line 54: Following on with comments on the PR in Registration app, this should be the default value, so we shouldn't need this here... 

I'd suggest instead, in gp.xml in config-pihemr, set this value to the null/empty and then here you should only need the two lines before in pih-config-haiti and pih-config-mirebalais where you set it to "paperrecord.patientDataDefinition.paperRecordIdentifier"

mogoodrich (changes requested):
Looks good, though we will also want a change in the man

mogoodrich (approved):
LGTM, thanks @cioan !

PR #76, opened 2021-04-15
UHM-5428 - Add tb screening result to status form, to show only if th…

mseaton:
…e result is not null


mogoodrich (approved):
LGTM... going to go ahead and merge since @mseaton  is on vacation...

PR #77, opened 2021-04-23
UHM-4354 Move Location, Location Tag, and Location Attribute Types installation out of PIH core and into config repos

brandones:
https://pihemr.atlassian.net/browse/UHM-4354


configuration/pih/pih-config-haiti-belladere.json
```diff
@@ -36,9 +36,7 @@
   },
 
   "locationTags": {
-    "Login Location": ["Belladère"],
-    "Medical Record Location": ["Belladère"],
-    "Visit Location": ["Belladère"]
```
mogoodrich, line 41: It's possible that Medical Record Location and Visit Location are two examples of ones that need to configured on a site-by-site basis, like Login, but lets go for moving them and see what happens.

Once moved, would be good to confirm that you can't start a visit at an unexpected location or create/request a paper record

mogoodrich (approved):
LGTM, with one comment... this is going to be a fun one to test! :)

brandones:
This change now uses a valid UUID for Cange. See the related liquibase change: https://github.com/PIH/openmrs-config-pihemr/pull/170

PR #81, opened 2021-05-11
UHM-5277: Adapt frontend directory to new system

brandones:
See https://github.com/PIH/openmrs-module-mirebalais/pull/153


mogoodrich (commented):
Same question as Mike... is "site" a new standard?

mseaton (approved):

PR #82, opened 2021-05-13
Update frontend config

brandones:


mogoodrich (approved):
I don't know much about this, but the change to use the variable openmrsBase seems correct!

PR #84, opened 2021-05-18
UHM-5528: Move encounter types into config

mseaton:


mogoodrich (commented):
Weren't there constants in constants.yml you pointed out as well that we should remove?

mseaton:
Good point.  I've pushed a follow-up commit to remove those.

PR #86, opened 2021-06-01
UHM-5565:  Application Role "archivistManager" and "labTechnician" do…

mseaton:
…n't appear to be getting installed after migration to Iniz install of Roles and Privleges


mogoodrich (approved):
LGTM, thanks @mseaton !

PR #101, opened 2021-08-31
Enabling SanteFanm and NCD on HSN

jmbabazi:


configuration/pih/pih-config-haiti-hsn.json
```diff
@@ -35,8 +37,8 @@
   },
 
   "locationTags": {
-    "Lab Results Location": ["COVID-19 | Izolman | HSN", "COVID-19 | UMI | HSN"],
-    "Login Location": ["COVID-19 | Izolman | HSN", "COVID-19 | UMI | HSN", "Sante Mantal | HSN"],
```
mogoodrich, line 39: @jmbabazi  we don't want to change the tags here.  HSN currently as 3 "sub-locations" (see attached image).

We should decide if these are still relevant and if we still want sub-locations at HSN... if so, we may need to add new locations for NCD and OB-GYN?

mogoodrich (changes requested):
@jmbabazi  we will need to figure out how we want to handle the locations at HSN... maybe we can discuss tomorrow after standup?

configuration/pih/pih-config-haiti-hsn.json
lnball: Do we want to remove COVID from HSN?  Maybe keep that around although they haven't been using...

PR #109, opened 2021-09-14
UHM-5404 - Example json configuration for adding the link to the Hiv …

mseaton:
…Emr On humci.

### DIFF
--- a/configuration/pih/pih-config-mirebalais-humci.json
+++ b/configuration/pih/pih-config-mirebalais-humci.json
@@ -75,6 +75,21 @@
     "matchUrl": "http://localhost:9000/match",
     "scanUrl": "https://localhost:9443/fingerprint/scan",
     "devicesUrl": "https://localhost:9443/fingerprint/devices"
-  }
+  },
+
+  "extensions": [
+    {
+      "id": "patientHeader.linkToHivEmr",
+      "type": "link",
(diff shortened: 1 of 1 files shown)


configuration/pih/pih-config-mirebalais-humci.json
```diff
@@ -75,6 +75,21 @@
...
+      "url": "http://localhost:8080/openmrs/pihcore/patient.page?patientId={{identifier}}&identifierType=${patientIdentifierType.hivemrV1.uuid}&dashboard=${program.hiv.uuid}",
+      "extensionPointId": "patientHeader.extraIdentifierLinks",
+      "label": "pihcore.hiv.viewPatientInHivEmr",
+      "icon": "icon-external-link",
+      "requiredPrivilege": null,
+      "extensionParams": {
+        "identifierType": "${patientIdentifierType.hivemrV1.uuid}"
+      }
+    }
+  ]
```
mogoodrich, line 93: I assume it was intentional to only set this up on HUM-CI for now?

mogoodrich (approved):
LGTM, assuming the intent was only to enable on HUM-CI for now/

configuration/pih/pih-config-mirebalais-humci.json
mogoodrich: Actually, probably should have this link to haitihivtest (instead of localhost) so we can test how this actually works?

mseaton: I'll hard-code that in for now, but generally feel like this is something we need to find a way to make come from an external source that can be set via puppet, maybe a runtime property.  I don't love having this hard coded here.

PR #114, opened 2021-09-17
UHM-5770: Refactor location tag configuration out of Iniz setup and i…

mogoodrich:
…nto PIH Config


mogoodrich:
Note that this comment cannot be merged in until the follow Iniz feature gets added:
https://github.com/mekomsolutions/openmrs-module-initializer/issues/137

cioan (approved):
Looks good!

configuration/pih/liquibase/liquibase-haiti.xml
```diff
@@ -0,0 +1,21 @@
...
+        <comment>
+            UHM-5770: Refactor location tag configuration into Iniz locationtagmaps
+            (One-time clear out of maps so they can be re-initiazed)
+        </comment>
+        <sql>
+            delete from location_tag_map;
+        </sql>
+    </changeSet>
+
+</databaseChangeLog>
```
mseaton, line 21: Per my comment on the other PR, can we avoid this?  It looks like you are defining a large superset of tags in each file, shouldn't this mean that iniz is deleting from location_tag_map as appropriate?  Are there other tags implicated?

mseaton (approved):
Looks great, nice work.  My only questions here and in the other PR is questioning the need for the liquibase cleanup step and how we add that in.

PR #117, opened 2021-09-28
UHM-5787:  Refactor the pih-config-haiti-hiv so that it inherits from…

mogoodrich:
… pih-config-haiti

### DIFF
--- a/configuration/pih/pih-config-haiti-hiv.json
+++ b/configuration/pih/pih-config-haiti-hiv.json
@@ -2,7 +2,6 @@
   "welcomeMessage": "Bienvenue dans l'EMR de VIH",
   "browserWarning": "Veuillez utiliser un navigateur <b>Google Chrome</b>",
 
-  "country": "HAITI",
   "site": "CROSS_SITE",
   "specialty": "HIV",
   "dossierIdentifierPrefix": "VIH",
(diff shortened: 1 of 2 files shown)


configuration/pih/pih-config-haiti.json
```diff
@@ -5,39 +5,6 @@
 
     "country": "HAITI",
 
-    "components": [
```
mogoodrich, line 8: Not sure why we had this, but there really isn't a need to set components in this top-level "Haiti" file, as we will aways be using it in conjunction with a country file like "pih-config-haiti-cange.json" or "pih-config-haiti-hiv.json"

configuration/pih/pih-config-haiti-hiv.json
```diff
@@ -61,44 +50,8 @@
...
-        "Telephone Number": "20"
-      }
-    },
-    "demographics": {
-      "mothersName": {
-        "required": true
-      }
-    }
+    "afterCreatedActions": [
+    ]
```
mogoodrich, line 54: The registration for the HIV server is exactly the same as the regular Haiti registration, except that we don't want to create a dossier number... if you look in the main pih-config-haiti it has a section:

`   "afterCreatedActions": [
          "bean:assignDossierNumber"
        ],`

This overrides that, so that no actions are triggered after a patient is created.

PR #118, opened 2021-09-29
UHM-5768 Transfer and referral separate

louidorjp:
tranferScreenshot
1-If the patient is transferred, it means that this patient was active in the site which will carry out this transfer. And this transfer can be internal (ie to another ZL site) and external (ie to another site that is not ZL).

2- If the patient is referred, it means that this patient was not active at this site. And it can be referenced by one of the ZL sites (internal) or by a site that is not from ZL (external).

So we need another section when the patient is referenced. because the transfer is different from the reference.
We need to create other concepts for referral in

### DIFF
--- a/configuration/pih/htmlforms/hiv/section-hiv-plan.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-plan.xml
@@ -7,6 +7,13 @@
         form fieldset {
             display: block;
         }
+        .side-by-side label {
+        display: inline-block;
+        }
+        form input[type="radio"], .form input[type="radio"] {
+        float: none;
+        display: inline-block;
+        }
 
         #test-orders input[type=checkbox],
         #id-treatment, #family-plan-section {
@@ -88,6 +95,11 @@
             width: 12em;
         }
     </style>
+    <script type="text/javascript">
+        jq(function() {
+            setUpExpandableTransferAndReferralDetails();
+        });
+    </script>
 
     <!-- Program enrollment (if the patient hasn't already been enrolled); use first ancestor location tagged as Program Location:
     note that we already do this on the history section, but add here case the history section is skipped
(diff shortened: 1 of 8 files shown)


mogoodrich (commented):
Thanks @louidorjp ... I just added some questions on the ticket, let's continue the discussion there.  I don't think we want to just directly change the text here.

louidorjp:
Screenshot from 2021-10-05 11-35-28
@mogoodrich please review this

configuration/pih/htmlforms/hiv/section-hiv-plan.xml
```diff
@@ -1168,47 +1180,48 @@
         </div>
     </section>
 
-    <section id="transfer-out-section" sectionTag="fieldset"
-             headerTag="legend" headerStyle="title" headerCode="pihcore.transfer.title">
+    <section id="transferIn" sectionTag="fieldset" headerTag="legend" headerStyle="title"
+             headerCode="pihcore.transferReferralIn">
         <div class="section-container-color">
-            <obsgroup groupingConceptId="PIH:13170">
```
mogoodrich, line 1174: This is the question about transferring/referring *out* on the plan sectoin of the HIV form.  I didn't realize we were doing transfer out as well as in.  

It's fine to do that, but we we shouldn't change the grouping concept here... I didn't realize that we wanted to use this new question for both transfers out and transfers in... I will update "PIH:13170" concept set to have the "referral or transfer" concept as well.

configuration/pih/htmlforms/hiv/section-hiv-plan.xml
```diff
@@ -1168,47 +1180,48 @@
...
+                         answerCodes="pihcore.referral,pihcore.transfer"
+                         answerSeparator="" toggle="transfer-referral-out-details"/>
                 </p>
-
-                <div id="transfer-out-details">
-                    <p class="small">
+                <div id="transfer-referral-out-details">
+                    <p class="side-by-side">
                         <obs id="transfer-out-spots"
-                             conceptId="PIH:Transfer out location" style="radio"
```
mogoodrich, line 1187: I don't think we want to change the concept here?  this is for the transfer out use caes, correct?

configuration/pih/htmlforms/hiv/section-hiv-plan.xml
```diff
@@ -1168,47 +1180,48 @@
...
-            <obsgroup groupingConceptId="PIH:13170">
+            <obsgroup groupingConceptId="PIH:13169">
                 <label>
-                    <uimessage code="pihcore.externalTransferQuestion" />
+                    <uimessage code="pihcore.externalTransferQuestion"/>
                 </label>
-                <p class="radio cb-align">
-                    <obs conceptId="PIH:HUM Disposition categories"
-                         answerConceptId="CIEL:160036" answerCode="emr.yes"
-                         style="checkbox" toggle="transfer-out-details"/>
```
mogoodrich, line 1181: I was thinking we would keep this question and just add in the second question, "referral or transfer" but you are right, it might be cleaning just to replace the questions... we will have to update the migration code to migration this information properly, however.

I don't know if a radio button will work though, because I think with a radio button once a user clicks on option they can't  unclick the question entirely... but maybe I'm wrong about that.

Thoughts?

mogoodrich (commented):
This generally looks good @louidorjp  but I saw that you added this question to the "transfer out" section of the HIV Plan and the "transfer in" of the PMTCT form, but don't we want to make the same change to the "Transfer In" in the section-history of the HIV form?

PR #125, opened 2021-10-05
Uhm 5750

louidorjp:
@mogoodrich
Debbie mentioned that the Screening date must validate.


mogoodrich (commented):
So adding the date validation here looks good, but it looks like the changes to the section-hiv-plan are still in this PR as well, and I though we were not going to make those changes/

PR #132, opened 2021-10-14
UHM-5817: validate format of Haiti phone number

mogoodrich:


configuration/pih/pih-config-haiti.json
```diff
@@ -70,6 +70,18 @@
           "mothersName": {
             "required": true
           }
+        },
+        "contactInfo": {
+            "phoneNumber": {
+              "regex": "(^\\d{8}$)|(^\\d{4}(?:\\)|[-|\\s])?\\s*?\\d{4}$)"
```
mogoodrich, line 76: I just combined the two regex from the zl.js.  I tested and it appears to work, but @louidorjp @rjacques89  would be good to review.

configuration/pih/pih-config-haiti.json
```diff
@@ -70,6 +70,18 @@
...
+            }
+        },
+        "contactPerson": {
```
mogoodrich, line 79: I had to add support in PIH Core for adding regex support "contactPerson", that change will be coming in another PR.

mseaton (approved):

PR #138, opened 2021-10-29
Uhm 5762

mseaton:
This PR moves in all of the jobs that were previously in the config-pihemr project, and reorganizes them a bit to enable us to take advantage of templating, iterations, and variable replacement, and to execute them against different source databases without repeating the same configurations over and over.

This also removes the old jobs and adds new jobs in that are geared towards 2 use cases:

1. Refresh all hiv / ovc / hivmigration tables from a single "HIV" server source
2. Refresh all remaining tables from all source databases (if no relevant data is in a source, it will contribute no rows)

This also introduces the notion of using maven dependencies and build process to copying config-pihemr reporting SQL files into the PETL project (so rather than accessing these by ../../../reports/.../foo.sql, you access them directly in the folder that you expect them to be in.


dmdesimone (approved):
Looks good!

jmbabazi (approved):
Looks good, so far we have this running on petl-tes. So merging it in

PR #146, opened 2021-11-19
UHM-5950:  Initial functional implementation of status widget that us…

mseaton:
…es a statusData configuration to retrieve and display particular data points about a patient.


configuration/pih/statusData/hiv/hivStatuses.yml
```diff
@@ -0,0 +1,7 @@
+# These status data definitions are what define the HIV STATUS widget on the HIV Program Dashboard
+
+- id: "nextMedicationPickupDate"
+  labelCode: "zl.status.medPickups.nextScheduledDate"
+  statusDataQuery: "nextMedicationPickupDate.sql"
+  valueExpression: "#if($data.numPickups == 0)$fn.translate('zl.status.medPickups.none')#{else}$fn.formatDate($data.nextPickupDate)#end"
+  formatExpression: "#if($data.numPickups == 0 || $fn.daysSince($data.nextPickupDate) > 28)status-data-alert#{else}#end"
```
mogoodrich, line 7: Same question, wondering how the formatting works...

configuration/pih/statusData/hiv/nextMedicationPickupDate.sql
```diff
@@ -0,0 +1,27 @@
+# Retrieve the  Next Medication Pickup Date (PIH:5096) from the last HIV drug dispensing encounter.
+#   If this date is more than 28 days ago, highlight it in red.
+#   If there have been no HIV Drug Dispensing encounters for this patient, display “No medication pickups” in red.
```
mogoodrich, line 3: Nit, but shouldn't these comments be in the hivStatuses.yml and the comment here should just be about what this query returns (number of med pickup encounters and most recent med pickup encounter, I believe)?

mogoodrich (approved):
LGTM, thanks @mseaton !

configuration/pih/statusData/hiv/hivStatuses.yml
mseaton: Yep, this needs documentation.  But yes, format translates to cssClass in the widget fragment (but could be used in other contexts as needed)

configuration/pih/statusData/hiv/nextMedicationPickupDate.sql
mseaton: Probably.  These are something I initially copied in here from the ticket in order to define what I was doing when I was writing the query.

mogoodrich (approved):
This is awesome, thanks @mseaton   

The detailed README is great... I did see your question about where it should live... config-pihemr is probably the best place for now?

PR #147, opened 2021-11-23
Uhm 5703 exposed infant data export initial code

dmdesimone:


configuration/pih/petl/jobs/refresh-hiv-data.yml
```diff
@@ -34,6 +34,7 @@ configuration:
           - tableName: "hiv_visit"
           - tableName: "ovc_program_encounters"
           - tableName: "hivmigration_data_warnings"
+          - tableName: "pmtct_infant_delivery"
```
mseaton, line 37: Just confirming - this is intended to run against the HIV system, and not the MCH system.  If so, this is the right file for this.  The order of the tableNames might be something you want to keep an eye on - keep them alphabetical or organized by data.  Putting this after the data warning table feels less well organized.

configuration/pih/petl/jobs/sql/extractions/pmtct_infant_delivery.sql
```diff
@@ -0,0 +1,123 @@
+set sql_safe_updates = 0;
+set session group_concat_max_len = 100000;
+
+select program_id into @eidProgram from program where uuid = '7e06bf82-9f1a-4218-b68f-823082ef519b';
+select encounter_type_id into @labResultsEnc from encounter_type where uuid = '4d77916a-0620-11e5-a6c0-1697f925ec7b';
+select name into @hivDispEncName from encounter_type where uuid = 'cc1720c9-3e4c-4fa8-a7ec-40eeaad1958c';
+
```
mseaton, line 7: Assuming you've double checked the uuids above.  You can use variables here like ${encounterType.HIV_DISPENSING.uuid}, but obviously that makes the SQL harder to test in a SQL editor

configuration/pih/petl/jobs/sql/extractions/pmtct_infant_delivery.sql
```diff
@@ -0,0 +1,123 @@
...
+-- current age, category
+update temp_eid t 
+set infant_age_in_months = current_age_in_months(patient_id);
+
+update temp_eid t 
+set age_cat_infant = 
+case 
+	when infant_age_in_months<2 then '<2 mo'
+	when infant_age_in_months>=2 and infant_age_in_months<=12 then '2-12 mo'
```
mseaton, line 55: Are you sure this is meant to be "<=12", and not "<12"?

configuration/pih/petl/jobs/sql/extractions/pmtct_infant_delivery.sql
```diff
@@ -0,0 +1,123 @@
...
+set pcr_test_date = obs_date(t.pcr_test_obs);
+
+update temp_eid t
+inner join obs o on o.obs_id = t.pcr_test_obs
+set pcr_result = concept_name(o.value_coded,@locale);
+
+-- art start date
+update temp_eid t
+set art_start_date = 
+	(select min(o.date_activated) from orders o
```
mseaton, line 82: For future-proof accuracy, I would do "min(ifnull(o.scheduled_date, o.date_activated))

configuration/pih/petl/jobs/sql/extractions/pmtct_infant_delivery.sql
```diff
@@ -0,0 +1,123 @@
...
+	where o.patient_id  = t.patient_id
+	and o.order_reason = concept_from_mapping('PIH','11197'));
+
+-- active on art
+update temp_eid t
+inner join orders o on o.patient_id = t.patient_id  and o.order_reason = concept_from_mapping('PIH','11197')
+	and o.date_activated < NOW() and o.date_stopped is null
```
mseaton, line 89: To check if active, I would:
* Use an inclusive condition, and also consider scheduled date for the start:
`ifnull(o.scheduled_date, o.date_activated) <= now()`
* Use an exclusive condition, and also consider a future auto expire date for the end:
`ifnull(o.date_stopped, o.auto_expire_date) is null or  > ifnull(o.date_stopped, o.auto_expire_date) > now()`

configuration/pih/petl/jobs/sql/extractions/pmtct_infant_delivery.sql
```diff
@@ -0,0 +1,123 @@
...
+set is_active_on_art = if(o.order_id is null, '0','1')
```
mseaton, line 90: If you are doing an inner join on orders with the above criteria, won't this always be "1"?

configuration/pih/petl/jobs/sql/extractions/pmtct_infant_delivery.sql
```diff
@@ -0,0 +1,123 @@
...
+inner join orders o on o.patient_id = t.patient_id  and o.order_reason = concept_from_mapping('PIH','11197')
+	and o.date_activated < NOW() and o.date_stopped is null
+set is_active_on_art = if(o.order_id is null, '0','1') 
+;
+-- is ltfu
+update temp_eid t 
+set next_dispensing_date = date(obs_value_datetime(latestEnc(patient_id, @hivDispEncName, null),'PIH','5096'));
+
+update temp_eid t 
+set is_LTFU = if(date_add(next_dispensing_date, interval 28 day) > CURRENT_DATE(),'0','1' );
```
mseaton, line 97: What's the difference between the CURRENT_DATE() used here, and the NOW() function used above?  Is one a date and the other a datetime?  Was this intentional?

mseaton (commented):
Generally looks good - a few comments.  I didn't review the concept ids and uuids and such.

configuration/pih/petl/jobs/sql/extractions/pmtct_infant_delivery.sql
dmdesimone: I was assuming that '<2 mo' means strictly less than 2 months, while '2-12 mo' means 2-12 inclusive (i.e. exactly 2 months to 12 months and 29 days or whatever).  Will highlight for Julia when she tests this.

configuration/pih/petl/jobs/sql/extractions/pmtct_infant_delivery.sql
dmdesimone: for when we schedule orders... good point.

configuration/pih/petl/jobs/sql/extractions/pmtct_infant_delivery.sql
dmdesimone: Yeah, that's the idea.  Let me know if if you think it should change.  (gonna commit like this)

PR #148, opened 2021-12-01
Uhm 5954 j9 etl for kellogg indicators

dmdesimone:


configuration/pih/petl/jobs/sql/extractions/mch_kellogg_data.sql
```diff
@@ -0,0 +1,333 @@
+-- set @locale = 'en';  -- uncomment for testing
+
+set sql_safe_updates = 0;
+set session group_concat_max_len = 100000;
+
```
mseaton, line 5: what is the purpose of the lines above?

configuration/pih/petl/jobs/sql/extractions/mch_kellogg_data.sql
```diff
@@ -0,0 +1,333 @@
...
+select program_id into @mchProgram from program where uuid = '41a2715e-8a14-11e8-9a94-a6cf71072f73';
+select concept_name(concept_from_mapping('PIH','YES'),@locale) into @yes;
+select concept_name(concept_from_mapping('PIH','NO'),@locale) into @no;
+select name into @socioEncName from encounter_type where uuid = 'de844e58-11e1-11e8-b642-0ed5f89f718b';
+select program_workflow_id into @mchWorkflow from program_workflow where uuid = '41a277d0-8a14-11e8-9a94-a6cf71072f73';
+set @past_med_finding = concept_from_mapping('PIH','10140');
+select encounter_type_id into @ob_gyn_enc_id from encounter_type where uuid = 'd83e98fd-dc7b-420f-aa3f-36f648b4483d';
```
mseaton, line 12: is there no encounter_type(uuid) function?

configuration/pih/petl/jobs/sql/schemas/mch_kellogg_data.sql
```diff
@@ -0,0 +1,40 @@
+CREATE TABLE mch_kellogg_data
```
mseaton, line 1: I really don't like calling this mch_kellogg_data.  Can we create a table with useful data that is not tied to a particular funder?  Can we just call this mch_j9_data?

configuration/pih/petl/jobs/sql/extractions/mch_kellogg_data.sql
```diff
@@ -0,0 +1,333 @@
...
+
+-- columns from socioeconomic form
+-- education level
+update temp_j9
+set education_level = latest_obs_value_coded(patient_id, 'CIEL','1712',@locale);
+
+-- able to read or write
+update temp_j9
+set able_read_write = 
+CASE latest_obs_value_coded(patient_id, 'CIEL','166855',@locale)
```
mseaton, line 89: This looks like the kind of thing that should be wrapped in a function and reused

mseaton (commented):
Made a few comments @dmdesimone .  Didn't really review the nuances or logic of the sql in detail, just some high-level thoughts.

configuration/pih/petl/jobs/sql/extractions/mch_kellogg_data.sql
dmdesimone: oh yeah...there is.  will use that.

configuration/pih/petl/jobs/sql/schemas/mch_kellogg_data.sql
dmdesimone: agreed.  will work with the reporting team to come up with something better.

configuration/pih/petl/jobs/sql/extractions/mch_kellogg_data.sql
dmdesimone: Yeah I know...don't know why I was resisting it.  I guess it would be a function that transforms a yes/no to boolean that I wrap around the existing function.

configuration/pih/petl/jobs/sql/extractions/mch_kellogg_data.sql
dmdesimone: I removed them.  If safe updates isn't turned off, it doesn't let you update a table without a where clause, which we don't need in this script.

And in some of our scripts (mostly exports, not ETLs) we have columns that exceed the default group_concat limit.  But again, don't need it for this script.

configuration/pih/petl/jobs/sql/extractions/mch_kellogg_data.sql
```diff
@@ -0,0 +1,311 @@
+-- set @locale = 'en';  -- uncomment for testing
```
mseaton, line 1: If the tablename is now "mch_j9_data" (which is much better), then this filename needs to change to "mch_j9_data.sql" as well.

configuration/pih/petl/jobs/sql/schemas/mch_kellogg_data.sql
```diff
@@ -0,0 +1,40 @@
+CREATE TABLE mch_j9_data
```
mseaton, line 1: Like above, this filename needs to change to "mch_j9_data.sql"

mseaton (commented):
Overall looks great @dmdesimone - just need to update the filenames and good to go.

PR #149, opened 2021-12-01
enabling components on cerca

jmbabazi:


mogoodrich (approved):
Great, @jmbabazi ... assuming these are all the components they want at Cerca, looks good to me!

PR #153, opened 2021-12-06
UHM-5918: Date Validation javascript function issues

mogoodrich:
(fix obs with obs dated function)

### DIFF
--- a/configuration/pih/htmlforms/hiv/section-hiv-state.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-state.xml
@@ -163,10 +163,10 @@
                     '<lookup expression="encounter.getEncounterDatetime().getTime()"/>',
                     '<uimessage code="pihcore.weeks"/>'
                 );
-            
+
                 setUpNextButtonForSections('hiv-state');
                 setUpDatepickerStartAndEndDateValidation('<uimessage code="pihcore.date.error.inTheFuture"/>','<uimessage code="pihcore.startDate.greater"/>');
-                setUpDatepickerInTheFutureValidation('<uimessage code="pihcore.date.error.inTheFuture"/>');
+                setUpObsWithObsDateTime('syph-checkbox');
             });
         </script>
     </ifMode>
@@ -934,9 +934,9 @@
 
             <div class="section-container">
                 <p>
-                    <obs conceptId="CIEL:164801" style="checkbox"
+                    <obs id="syph-checkbox" conceptId="CIEL:164801" style="checkbox"
                          answerConceptId="CIEL:1065" answerCode="pihcore.screeningSyph"
-                         dateLabel="Date" toggle="syph-screen" class="dateDatepickerInTheFuture"/>
+                         dateLabel="Date" toggle="syph-screen" />
                 </p>
                 <p id="syph-screen" class="side-by-side">
                     <label>
--- a/configuration/pih/scripts/global/zl.js
+++ b/configuration/pih/scripts/global/zl.js
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
     jq("#submit,.confirm").prop("disabled", val);
 }
 
-function setUpDatepickerInTheFutureValidation(badDateInTheFutureMsg){
+/**
+ * This function provides specific client-side functionality for with "checkbox" style obs with an obs date component;
+ * It is currently used for the "screening for syphilis", and configured by applying a class "dateDatepickerInTheFuture" to the relevant obs
+ *
+ * It does two main things:
+ *   ** Hide/show the datepicker input based on whether the checkbox is checked
+ *   ** Doesn't allow date selected to be ahead of the current date
+ *
+ * Ideally, 1) The HTML Form Entry module would handle this validation client-side (currently it only handles it
+ * server-side) and 2) would not allow the date to after the *encounter date* (not just the current date)
+ *
+ * We have a ticket for the above work, see: https://issues.openmrs.org/browse/HTML-799 , and if we implement this
+ * we could potentially rework this function to just be about hiding/showing the date
+ *
+ * @param badDateInTheFutureMsg
+ */
+function setUpObsWithObsDateTime(widgetId){
+    getField(widgetId + '.date').hide();
+    getField(widgetId + '.date').datepicker('option', 'maxDate', new Date());
 
-    jq('.dateDatepickerInTheFuture').each(function(j, domEl){
-        jq(this).find('input[type=text]').hide();
-        jq(this).change(function (e) {
-            const  date=Date.parse(jq(this).find('input[type=text]').val())
-            const  isCheched=Date.parse(jq(this).find('input:checked').val())
-            if(isCheched) {
-                if (date > Date.now()) {
-                    jq(this).find('span').show();
-                    jq(this).find('span').text(badDateInTheFutureMsg);
-                    setButtonsDisabled(true)
-                } else {
-                    jq(this).find('span').hide();
-                    jq(this).find('span').text('');
-                    setButtonsDisabled(false)
-                }
-                jq(this).find('input[type=text]').show();
-            }else {
-                jq(this).find('input[type=text]').prop('value', '')
-                jq(this).find('span').text('');
-                jq(this).find('input[type=text]').hide();
-                setButtonsDisabled(false)
-            }
-        })
-    });
-
+    getField(widgetId + '.value').change(function() {
+      const isChecked = getValue(widgetId + '.value');
+      if(isChecked) {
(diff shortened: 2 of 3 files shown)


mogoodrich:
So I did some refactoring here, I'll add some comments to the PR, but mainly I used main of the JavaScript helper functions provided by HTML Form Entry for interacting with the fields created by HTML Form Entry.  Louidor, my apologies, I should have posted a link to the documentation for them earlier, they are here:

Basically, if you add an id attribute an HTML Form Entry element, it allows you to access it (and bind listeners to it) via "getField" and "getValue" functions, and to "setValue"

Note that this PR is currently only for the validator that validates that the date is in the future... I'm having some issues with the functionality that binds a listener to a datepicker widget rendered by HTML Form Entry, which appears to be something that worked previously but is not working now, and would be required for the "start date" and "end date" validator to work.

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -163,10 +163,10 @@
                     '<lookup expression="encounter.getEncounterDatetime().getTime()"/>',
                     '<uimessage code="pihcore.weeks"/>'
                 );
-            
+
                 setUpNextButtonForSections('hiv-state');
                 setUpDatepickerStartAndEndDateValidation('<uimessage code="pihcore.date.error.inTheFuture"/>','<uimessage code="pihcore.startDate.greater"/>');
-                setUpDatepickerInTheFutureValidation('<uimessage code="pihcore.date.error.inTheFuture"/>');
```
mogoodrich, line 169: I changed this method so that instead of setting up one or more validators of this type, it sets it up only for the one referenced by the passed-in ID, in this case "syph-checkbox"... if we were using this same validation is multiple places on the form, we just call this method twice with the two different IDs.

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -934,9 +934,9 @@
 
             <div class="section-container">
                 <p>
-                    <obs conceptId="CIEL:164801" style="checkbox"
+                    <obs id="syph-checkbox" conceptId="CIEL:164801" style="checkbox"
```
mogoodrich, line 937: I added an id in so that we can access it via the HTML Form Entry helper functoins, see below...

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -934,9 +934,9 @@
...
                          answerConceptId="CIEL:1065" answerCode="pihcore.screeningSyph"
-                         dateLabel="Date" toggle="syph-screen" class="dateDatepickerInTheFuture"/>
+                         dateLabel="Date" toggle="syph-screen" />
```
mogoodrich, line 939: I removed the class here as I'm no longer using it.

configuration/pih/scripts/global/zl.js
```diff
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
...
+ *
+ * Ideally, 1) The HTML Form Entry module would handle this validation client-side (currently it only handles it
+ * server-side) and 2) would not allow the date to after the *encounter date* (not just the current date)
+ *
+ * We have a ticket for the above work, see: https://issues.openmrs.org/browse/HTML-799 , and if we implement this
+ * we could potentially rework this function to just be about hiding/showing the date
+ *
+ * @param badDateInTheFutureMsg
+ */
+function setUpObsWithObsDateTime(widgetId){
```
mogoodrich, line 218: I changed the function name to make it closer to what it was actually doing, see my description above.

configuration/pih/scripts/global/zl.js
```diff
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
...
+    getField(widgetId + '.date').hide();
```
mogoodrich, line 219: This fetches the field associated with the date widget obs associated with the "syph-checkbox" and hides it.  "getField" will return a jQuery object.

configuration/pih/scripts/global/zl.js
```diff
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
...
+    getField(widgetId + '.date').datepicker('option', 'maxDate', new Date());
```
mogoodrich, line 220: I changed this so instead of validating the date "on change", it uses the datepicker syntax to update the datepicker so that the date can't be after today... this is the preferred approach, I think, but I also did it because I couldn't bind a change listener to the datepicker properly (this used to work, I believe).

configuration/pih/scripts/global/zl.js
```diff
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
...
+
+    getField(widgetId + '.value').change(function() {
```
mogoodrich, line 222: This gets the field associated with the "value" widget of the "syph-checkbox" obs, which is the checkbox, and allows us to bind a jQuery change handler to it.

configuration/pih/scripts/global/zl.js
```diff
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
...
+      const isChecked = getValue(widgetId + '.value');
```
mogoodrich, line 223: This gets the value of the same widget.

configuration/pih/scripts/global/zl.js
```diff
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
...
+      if(isChecked) {
+        getField(widgetId + '.date').show();
+      }
+      else {
+        setValue(widgetId + '.date','')
```
mogoodrich, line 228: Note that there's also a "setValue" method as well (that we use here to clear out the date widget when we "hide" it)

configuration/pih/scripts/global/zl.js
```diff
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
...
+ *
+ * It does two main things:
+ *   ** Hide/show the datepicker input based on whether the checkbox is checked
+ *   ** Doesn't allow date selected to be ahead of the current date
+ *
+ * Ideally, 1) The HTML Form Entry module would handle this validation client-side (currently it only handles it
+ * server-side) and 2) would not allow the date to after the *encounter date* (not just the current date)
+ *
+ * We have a ticket for the above work, see: https://issues.openmrs.org/browse/HTML-799 , and if we implement this
+ * we could potentially rework this function to just be about hiding/showing the date
```
mogoodrich, line 214: Note all my comments/thoughts above.

configuration/pih/scripts/global/zl.js
```diff
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
...
-                jq(this).find('span').text('');
-                jq(this).find('input[type=text]').hide();
-                setButtonsDisabled(false)
-            }
-        })
-    });
-
+/**
+ * This function provides specific client-side functionality for with "checkbox" style obs with an obs date component;
+ * It is currently used for the "screening for syphilis", and configured by applying a class "dateDatepickerInTheFuture" to the relevant obs
```
mogoodrich, line 204:
```suggestion
 * It is currently used for the "screening for syphilis".
```

mseaton (approved):
Looks good to me @mogoodrich .  If we are doing a lot of "getField(...)" with the same concatenated field names several times, might be worth doing it once at the top as a variable and then reusing, but overall looks good.

mogoodrich:
Thanks @mseaton !  Merging this in to keep the sprint moving, but post-commit questions/reviews welcome @louidorjp  and @brandones !

configuration/pih/scripts/global/zl.js
```diff
@@ -199,31 +199,34 @@ function setButtonsDisabled(val){
...
+ *   ** Hide/show the datepicker input based on whether the checkbox is checked
+ *   ** Doesn't allow date selected to be ahead of the current date
+ *
+ * Ideally, 1) The HTML Form Entry module would handle this validation client-side (currently it only handles it
+ * server-side) and 2) would not allow the date to after the *encounter date* (not just the current date)
+ *
+ * We have a ticket for the above work, see: https://issues.openmrs.org/browse/HTML-799 , and if we implement this
+ * we could potentially rework this function to just be about hiding/showing the date
+ *
+ * @param badDateInTheFutureMsg
```
brandones, line 216: Should update this param name in the docstring

brandones:
Looks good aside from one nit, thanks @mogoodrich !

configuration/pih/scripts/global/zl.js
mogoodrich: Done!

PR #154, opened 2021-12-07
UHM-5973 Added B6 (vitamin) to drug order and hiv dispensing

lnball:
At this late date, I don't want to make changes to the HIV forms without review.

### DIFF
--- a/configuration/pih/htmlforms/hiv/hiv-dispensing.xml
+++ b/configuration/pih/htmlforms/hiv/hiv-dispensing.xml
@@ -712,6 +712,27 @@
                         </td>
                     </tr>
                 </obsgroup>
+
+                <!-- Vitamin B6 -->
+                <obsgroup groupingConceptId="CIEL:1442" >
+                    <tr>
+                        <td>
+                            <uimessage code="B6" />:
+                            <span class="small hidden">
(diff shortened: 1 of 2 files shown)


lnball:
@mogoodrich @mseaton 

This is in the ticket:

- Add to prophylaxis drug order
-- Added to order – just after INH in the list
-- Did not add to the HIV plan V3.

- Add to HIV dispensing form
-- Hidden field with medication type = prophylaxis.  This seemed better than medication type of INH, but please advise
-- At least one drug should be selected on HIV dispense from, but not including B6 since it should always be with INH.
-- Not translating because B6 is same in english/french/kreyol.
-- Do we want to add B6 to the HIV dispensing V1 form?

mseaton (approved):

mseaton:
@lnball I approved this but that was before I saw your comments.

* No, we don't want to add this to the v1, unless we collected it in v1.  We didn't, did we?  This is new data.
* I don't know the actual reason for the B6.  If it's part of a prophylaxis regimen, then I guess "Prophylaxis" is appropriate.  But we should also be thinking about what order reasons we want to support, and why - i.e. where are these used.  @dmdesimone  has started using reasons in reports to determine things like the patient's ART start date.  So we need to think through the data use implications of associating reasons  Reasons are also not required in the data model.

mogoodrich:
Looks like @mseaton  has this covered, but if you want me to give it a review @lnball as well lmk...

PR #155, opened 2021-12-07
UHM-5972: Fix start and end dates for syphilis on PMTCT intake and fo…

mogoodrich:
…llowup

### DIFF
--- a/configuration/pih/scripts/global/zl.js
+++ b/configuration/pih/scripts/global/zl.js
@@ -88,69 +88,33 @@ function setUpExpandableContacts(badPhoneNumberMsg) {
 
 }
 
-function setUpDatepickerStartAndEndDateValidation(badDateInTheFutureMsg,badStartDateGreaterThanEndDateMsg){
+/**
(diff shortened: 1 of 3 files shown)


configuration/pih/scripts/global/zl.js
```diff
@@ -216,17 +181,18 @@ function setButtonsDisabled(val){
...
-      const isChecked = getValue(widgetId + '.value');
-      if(isChecked) {
-        getField(widgetId + '.date').show();
-      }
-      else {
-        setValue(widgetId + '.date','')
-        getField(widgetId + '.date').hide();
-      }
-    })
+    if (getField(widgetId + '.date') && getField(widgetId + '.value')) {
```
mogoodrich, line 178: this is a follow-up to the last commit, to make this null-safe... we really shouldn't be initializing this unless we know the widgets exist, but are some cases where the syphilis question isn't shown on the form but the setup is still happening.

mseaton (approved):
This looks great @mogoodrich .  Nice work!

mogoodrich:
Thanks @mseaton , merging!

PR #156, opened 2021-12-08
UHM-5972: code refactoring for Fix start and end dates for syphilis

louidorjp:
@mogoodrich please review this PR

### DIFF
--- a/configuration/pih/scripts/global/zl.js
+++ b/configuration/pih/scripts/global/zl.js
@@ -93,28 +93,31 @@ function setUpExpandableContacts(badPhoneNumberMsg) {
  * and that the "end date" cannot be before the "start date" by updating the min and max dates of the respective
  * datepickers
  */
-function setUpDatepickerStartAndEndDateValidation(startDateWidgetId, endDateWidgetId){
+function setUpDatepickerStartAndEndDateValidation(){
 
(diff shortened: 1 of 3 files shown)


configuration/pih/scripts/global/zl.js
```diff
@@ -93,28 +93,31 @@ function setUpExpandableContacts(badPhoneNumberMsg) {
...
-      startDatepicker.change(function () {
-        let startDate = getField(startDateWidgetId + '.value').datepicker('getDate');
-        if (startDate) {
-          endDatepicker.datepicker('option', 'minDate', startDate);
+        const startDatepicker = jq(jq(this).find('.startDate'));
+        const endDatepicker = jq(jq(this).find('.endDate'));
+
+        if (startDatepicker) {
+            startDatepicker.change(function () {
+                let startDate = startDatepicker.find('input[type=text]').datepicker('getDate');
```
mogoodrich, line 105: there's a risk here that if the implementation of the HTML Form Entry datepicker changes, and doesn't have an underlying element input of type=text, this might break, but I agree that is a small risk.

mogoodrich (approved):
Thanks @louidorjp ... I added one small comment, but generally looks good, I will merge this in... Dave was going to be testing shortly, but I asked him to wait until this change goes through.

PR #157, opened 2021-12-08
Removing non ncd components and enabling lab result form

jmbabazi:


configuration/pih/pih-config-haiti-cerca.json
```diff
@@ -27,7 +24,8 @@
     "todaysVisits",
     "visitManagement",
     "visitNote",
-    "uhmVitals"
+    "uhmVitals",
```
mogoodrich, line 27: there are two different vitals components, confusingly one named "uhmVitals" and one names "vitals"... not sure which one you want to use, but the differences are examples in this ticket (which is about refactoring it and making it cleaner)

https://pihemr.atlassian.net/browse/UHM-5790

configuration/pih/pih-config-haiti-cerca.json
```diff
@@ -27,7 +24,8 @@
...
+    "labResults"
```
mogoodrich, line 28: Nit, but can we move this up so it is alphabetized?  it makes it easier to do a comparison between the config files for two different sites.

mogoodrich (commented):
I made one small comment and addressed your question around "uhmVitals", but generally looks good!

configuration/pih/pih-config-haiti-cerca.json
jmbabazi: Thanks this is alphabetized.

configuration/pih/pih-config-haiti-cerca.json
jmbabazi: For now we are going with uhmVitals

PR #159, opened 2021-12-10
Uhm 5968

mseaton:


mogoodrich (approved):
I didn't review the actual logic in any detail, defer to @dmdesimone  for that... but so sweet to see us being able to add statuses almost entirely using code...  @mseaton  thanks so much for building this out.

PR #160, opened 2021-12-10
UHM-6014 Changed concept for INH prophylaxis treatment line; UHM-6015…

lnball:
In addition to switching the concept for INH prophylaxis treatment line, also (UHM-6015) Fixed Behavior layout on hiv intake history.  

@mseaton As you mentioned, the migration script will need to be changed.


mseaton (approved):
Looks good, let's make sure we create tickets for the migration and also make sure the new concept is added to the config before we merge this in...

PR #161, opened 2021-12-10
fix name of cerca location maps file

mogoodrich:


jmbabazi:
thanks @mogoodrich , merging this in.

PR #162, opened 2021-12-11
Uhm 4342

mseaton:


mseaton:
@mogoodrich it would be helpful if you could confirm these configurations based on what is actually running on each of these servers, and whether any tweaks are needed to reflect where we want to set specific values.

mogoodrich (approved):
Confirmed this is the case on Haiti HIV Test, HUM Demo, ci.pih-emr.org and ami.pih-emr.org...

Unfortunately, HUM-CI is a bit different and may need some custom tweaks, as we had to set up a second identifier generator there using "X" because the first one ran out of identifiers...   fine... we will probably need to tweak something with HUM-CI when merging this in, but I'm fine with a hack...

PR #163, opened 2021-12-13
UHM-5959 - Next Dispensing date and Next Appointment Date should be v…

mseaton:
…alidated against encounter date


mogoodrich (approved):
LGTM, thanks @mseaton

PR #175, opened 2022-01-11
table that provides visibility into how old the database restored was

jmbabazi:


mogoodrich (approved):
LGTM to me, thanks @jmbabazi !

PR #176, opened 2022-01-12
remove error handling from HUM-CI and OpenMRS jobs

mogoodrich:


mogoodrich:
I'm proposing removing the error handling from all the config-zl PETL jobs for now, just so that we have a "fail fast" scenario while we are debugging the jobs themselves... the retries add noise and seem to make it harder to track down the actual errors.  Does this work for you guys?

jmbabazi (approved):
Looks good @mogoodrich

mseaton (approved):
Yes, 100%.  Not sure why we had this in there given these servers should always be available.  A holdover from demoing this feature that we never removed.  That should definitely help.

mogoodrich:
Thanks, merged!

PR #184, opened 2022-01-24
current enrollment location for hiv_patient_table

jmbabazi:


configuration/pih/petl/jobs/sql/extractions/hiv_patient.sql
```diff
@@ -31,7 +31,7 @@ CREATE TABLE temp_patient
     birthplace_province         VARCHAR(100),
     initial_health_center       VARCHAR(100),
```
mseaton, line 32: Don't you need to change this column name to "initial_enrollment_location"?

mseaton (changes requested):

PR #186, opened 2022-01-26
additiob columns to dispensing table

jmbabazi:


mseaton (approved):
Looks fine in principle, I didn't look at every detail.  Will need to do testing to ensure that all of the right concepts are used etc.  It's a lot of duplicated code - is there no function that exists for this?

PR #189, opened 2022-02-01
UHM-6108: HIV Medication dispensing form v1.0 does not save meds in t…

mogoodrich:
…he correct order


mseaton (approved):
I really don't understand the code behind all of this, but approving this PR and the one in HFE, as you appear to have a solid grasp of it and have a clear use case that you have to test it here.

cioan (approved):
Thanks @mogoodrich !

mogoodrich:
Thanks guys, merging this in!

PR #191, opened 2022-02-09
UHM-6027 - Add missing french translations

mseaton:


lnball:
Why add to the ZL translation files when this should likely be part of messages_fr.properties?

PR #192, opened 2022-02-14
Enabling j9 on hinche-ci

jmbabazi:


configuration/pih/pih-config-hinche-ci.json
```diff
@@ -0,0 +1,9 @@
+{
+  "components": [
+    "homeVisitsOnClinicianDashboard",
+    "j9"
+  ],
```
mogoodrich, line 5: Unfortuntely we don't have support yet for merging at the component level... ie, the components defined here will override all that are defined in pih-config-hinche, so you'll need to include the full list of components defined in pih-config-hinche.json here, plus these two new ones...

mogoodrich (changes requested):

mogoodrich (approved):
LGTM!

For reference, you only have to include variables that you want to change... so, for instance if "welcomeMessage" and "site"  are the same as in pih-config-hinche.json, you don't need to add them here.

The issue is, once you *do* want to override a variable, like "components", you need to define it fully in the new file.  Hope that makes sense, happy to discuss further!

PR #198, opened 2022-04-08
UHM-6402 Updated order reasons on HIV forms (plan section)

lnball:
There are some new concepts for the drug order reasons per discussion with @dmdesimone today.

See https://pihemr.atlassian.net/browse/UHM-6402 for the concepts.

In the past, there were NO order reasons for HIV drugs shown on the web UI.  It was hidden but always set HIV disease as the reason.  Now there are 3 possible reasons so there is a dropdown with these choices:
* HIV treatment
* HIV prophylaxis
* PMTCT


mogoodrich (approved):
LGTM, but defer more to @mseaton and @dmdesimone  as I haven't been following this one too closely.

PR #199, opened 2022-04-22
UHM-6363: Configure (and release) link from local EMRs to HIV Cloud

mogoodrich:


mseaton (approved):
Well, I guess it's OK, but not really maintainable like this.  I guarantee someone will forget to update all of these at some point if they need to make changes.  Why don't we just add it to the CALF wrapped in a component?

PR #200, opened 2022-04-26
UHM-6363: Configure (and release) link from local EMRs to HIV Cloud

mogoodrich:


mogoodrich:
New PR... since I'm no longer using a component, just the presence of the runtime property, all the configuration will happen in Puppet, just removing the extension config here.

mseaton (approved):

PR #203, opened 2022-05-19
HAI-668:Add DTG 5mg to the list of drugs

louidorjp:
Please review this PR @lnball @rjacques89 @mogoodrich

### DIFF
--- a/configuration/pih/htmlforms/hiv/hiv-dispensing.xml
+++ b/configuration/pih/htmlforms/hiv/hiv-dispensing.xml
@@ -597,6 +597,8 @@
                                 78f96a76-dfbe-11e9-8a34-2a2ae2dbcce4,
 
                                 a6985b15-5fd6-4bdf-93f6-62930a438464,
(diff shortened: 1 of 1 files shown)


configuration/pih/htmlforms/hiv/hiv-dispensing.xml
```diff
@@ -615,6 +617,8 @@
                                 TDF/3TC 300/300mg,
 
                                 ABC/3TC 120/60mg,
+                                DTG 50mg,
+                                DTG 5mg,
```
lnball, line 621: @louidorjp Looks good, but I would suggest changing the order so it's 50, 10, 5mg (instead of 50, 5, 10) and same order as the ARV #2

lnball (changes requested):
Small change but looks good and nice improvement.

louidorjp:
@lnball list of drugs ARV #1 updated

lnball:
Looks good.  Thanks @louidorjp

PR #208, opened 2022-06-29
Hai 692 possibility to add a foreign address in the sites that are closed to the dominican border

ljoedavessenn:
Adding Dominican Republic provinces for address option.
Update phone number regex for international number.
Add phone number regex in mirebalais config.


configuration/addresshierarchy/haiti_address_hierarchy_entries.csv
```diff
@@ -27283,3 +27283,35 @@ Haiti|Sud-Est|Jacmel|6ème Montagne La Voute^211-06|Ville de Jacmel
...
+République Dominicaine|Samaná|-|-|-
+République Dominicaine|Sánchez Ramírez|-|-|-
+République Dominicaine|San Cristóbal|-|-|-
+République Dominicaine|San José de Ocoa|-|-|-
+République Dominicaine|San Juan|-|-|-
+République Dominicaine|San Pedro de Macorís|-|-|-
+République Dominicaine|Santiago|-|-|-
+République Dominicaine|Santiago Rodríguez|-|-|-
+République Dominicaine|Santo Domingo|-|-|-
+République Dominicaine|Valverde|-|-|-
```
lnball, line 27317: Does 'Valverde' work?  Do you need a newline?

configuration/pih/pih-config-mirebalais.json
```diff
@@ -70,6 +70,17 @@
       "mothersName": {
         "required": true
       }
+    },
+    "contactInfo": {
+      "phoneNumber": {
+        "regex": "(^\\d{8}$)|(^\\d{4}(?:[-| ])?\\d{4}$)|(^\\+?(?:\\d ?){7,14}\\d$)"
```
lnball, line 76: The phone number format would be a change for EVERY ZL EMR – including HUM registration.  If you want to do this, I recommend that you notify HUM registration clerks ( can help).  I recommend you provide format tip (ie. xxxx-xxxx) with the label.

Otherwise when we do the next software upgrade at HUM, it will be difficult for users who do not use that format.

Today they can try anything into the telephone number:

lnball (changes requested):
Please change on the 2 comments:  new line AND change for HUM production

configuration/pih/pih-config-mirebalais.json
ljoedavessenn: So I'd better do the two changes in different branches so that I could create a different pull request for the phone number validation?

PR #209, opened 2022-07-01
Hai 708 update phone number validation so that it allows international phone

ljoedavessenn:
Update regex for international number
Add phone number format to label


lnball (approved):
LGTM.  Nice work.

Be sure to discuss with Bernard and HUM registration data clerks.

PR #210, opened 2022-07-04
HAI-697: Add validation to HIV return visit date

louidorjp:
Please review this PR
@lnball

### DIFF
--- a/configuration/pih/htmlforms/hiv/section-hiv-plan.xml
+++ b/configuration/pih/htmlforms/hiv/section-hiv-plan.xml
@@ -96,6 +96,7 @@
         }
     </style>
 
+
     <!-- Program enrollment (if the patient hasn't already been enrolled); use first ancestor location tagged as Program Location:
     note that we already do this on the history section, but add here case the history section is skipped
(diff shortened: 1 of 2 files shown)


lnball:
@louidorjp I think your code restricts "return visit date" to be within 12 months of the current date.  Is that correct?

This could be a problem if a form is back entered.  For example, the current date is "11 Jul 2022" but the encounter date for the HIV form is "1 Jan 2020" and the return visit date is "1 July 2020".  Will an error message appear?  

I think you want to use encounter date instead of current date for your calculations.

mogoodrich:
Hey @louidorjp , how are you?

So the the

lnball:
FYI @mogoodrich.  Spoke to @louidorjp.  He figured out the code and will push fixes to the branch.  We'll review.

mogoodrich:
oh, thanks @lnball @louidorjp ... I got distracted, but I'm in the midst of doing a review, will go ahead and finish that anyway!

mogoodrich (approved):
Thanks @louidorjp !  There are a few things that I think will need to be fixed, but I think it will easier for me to comment on them via a follow-up review... I also want to merge so I can test some of my changes, so I will test merge this now.

FYI, there are some utility JavaScript functions provided by HTML Form Entry that can come in helpful when writing this code:

PR #211, opened 2022-07-13
HAI-697: Add validation to HIV return visit date

mogoodrich:
Thanks @louidorjp  and @lnball ! 

Attached is a PR for HAI-697 based on the original PR merged yesterday from @louidorjp 

We have some JavaScript utility functions built in to HTML Form Entry that can be used for validation (and other things) that we haven't done a good job of highlighting.  The documentation can be found here:

https://wiki.openmrs.org/display/docs/HTML%2BForm%2BEntry%2BJavaScript%2BReference

I've made some suggested changes to use these functions in the PR and will add some comments in the PR explaining their usage.

Also, I found a few issues:

* If there are multiple validations on the same form and one "passes", it could override another failure (I'll add more details and what I believe should fix it to the PR)
* This may only be a "nitpick" but I noticed that the comparison only happened on the "month" level... for instance, if the encounter is on Sept 22, 2022, return appointments are allowed up through Sept 30, 2023 (not Sept 22, 2023) I defer to you all if this is okay.
* Message codes are appearing as codes, but I think this is just because the PR 

I've done some testing with my changes, but certainly should be tested again once merged in, as I certainly could have missed something.

Will add some comments to the PR now and then flag for you all to review!

### DIFF
--- a/configuration/pih/scripts/global/zl.js
+++ b/configuration/pih/scripts/global/zl.js
@@ -194,17 +194,17 @@ function setupReturnVisitDateValidation(encounterDate,returnVisitDateLessThanEnc
 
    const returnVisitDateValidation = function() {
      const nextVisitDate = domEl.datepicker('getDate');
-     const months = (nextVisitDate.getFullYear() - encounterDate.getFullYear()) * 12;
+     const differnenceInYears = nextVisitDate.getFullYear() - encounterDate.getFullYear();
 
-     if (months < 12  && nextVisitDate > encounterDate) {
+     if (differnenceInYears < 1  && nextVisitDate > encounterDate) {
        getField('apptDate.error').text('').hide();
        return true;
      }
-     else if(months == 12 && nextVisitDate.getMonth() <= encounterDate.getMonth()) {
+     else if(differnenceInYears == 1 && nextVisitDate.getMonth() <= encounterDate.getMonth()) {
        getField('apptDate.error').text('').hide();
        return true;
      }
-     else if(months <= 0 && encounterDate > nextVisitDate) {
+     else if(differnenceInYears <= 0 && encounterDate > nextVisitDate) {
        getField('apptDate.error').text(returnVisitDateLessThanEncounterDateMsg).show();
        return false;
      }


configuration/pih/scripts/global/zl.js
```diff
@@ -185,32 +185,35 @@ function setUpObsWithObsDateTime(widgetId){
...
-      jq('.submit').prop("disabled", true);
-    }
-
-  })        
+   domEl.datepicker('option','yearRange', yrRange);
+   domEl.prop("readonly", "readonly");
+
+   const returnVisitDateValidation = function() {
+     const nextVisitDate = domEl.datepicker('getDate');
+     const months = (nextVisitDate.getFullYear() - encounterDate.getFullYear()) * 12;
```
mogoodrich, line 197: This is a "nitpick" and I didn't change it, but I think this variable would be clearer if you *didn't* multiple it by 12 and just called it "differenceInYears" or something like that... it took me a little bit to understand the validation logic below and I think something like this would have been clearer:

`if (differnenceInYears < 1  && nextVisitDate > encounterDate)`

("if the two years are the same, just validate that return visit date is after encounter date'")

configuration/pih/scripts/global/zl.js
```diff
@@ -185,32 +185,35 @@ function setUpObsWithObsDateTime(widgetId){
...
+
+     if (months < 12  && nextVisitDate > encounterDate) {
+       getField('apptDate.error').text('').hide();
+       return true;
```
mogoodrich, line 201: Note that I replaced the code that disables/enabled the submit with one that returns true or false.  I'll explain the true/false below, but the main problem with disabling the submit here is: what if there's another validatoin function on this page? If a user enters an invalid value for appointment date, but then enters some valid data for the other field, the submit button will be re-enabled even though the appt date is still invalid.

We've added some custom functions to HTML Form Entry to help us get around this problem, see below.

configuration/pih/scripts/global/zl.js
```diff
@@ -185,32 +185,35 @@ function setUpObsWithObsDateTime(widgetId){
...
+       getField('apptDate.error').text(returnVisitDateLessThanEncounterDateMsg).show();
+       return false;
+     }
+     else {
+       getField('apptDate.error').text(badReturnVisitDateMsg).show();
+       return false;
+     }
+   }
+
+    jq(domEl).change(returnVisitDateValidation);
```
mogoodrich, line 217: This sets up the change handler just like it did in the old code, except now it *only* show/hides the error message, it doesn't disable/enable the submit button.

(One thing I'm not 100% sure of... does anyone know if having the function return true/false affects the way the change handler works? If so, this may be a problem).

configuration/pih/scripts/global/zl.js
```diff
@@ -185,32 +185,35 @@ function setUpObsWithObsDateTime(widgetId){
...
-  })        
+   domEl.datepicker('option','yearRange', yrRange);
+   domEl.prop("readonly", "readonly");
+
+   const returnVisitDateValidation = function() {
+     const nextVisitDate = domEl.datepicker('getDate');
+     const months = (nextVisitDate.getFullYear() - encounterDate.getFullYear()) * 12;
+
+     if (months < 12  && nextVisitDate > encounterDate) {
+       getField('apptDate.error').text('').hide();
```
mogoodrich, line 200: This uses the "getField" method (documentation in the link I added above) to hide/set the error text on the "span" with class "error" that HTML Form Entry provides for us.

configuration/pih/scripts/global/zl.js
```diff
@@ -185,32 +185,35 @@ function setUpObsWithObsDateTime(widgetId){
...
+       return false;
+     }
+     else {
+       getField('apptDate.error').text(badReturnVisitDateMsg).show();
+       return false;
+     }
+   }
+
+    jq(domEl).change(returnVisitDateValidation);
+    beforeSubmit.push(returnVisitDateValidation);
```
mogoodrich, line 218: This added this same "returnVisitDateValidation" method to a list of functions that are called when the user tries to submit a form.  This assures that the validation will be checked when the user attempts to save, and works around the issue I described above when/if there are multiple fields that need to validated on a page.

You can see the documentation for how this function works here:
https://wiki.openmrs.org/display/docs/HTML%2BForm%2BEntry%2BJavaScript%2BReference#HTMLFormEntryJavaScriptReference-Definingjavascriptfunctionstocallbeforeformvalidationandsubmission

configuration/pih/scripts/global/zl.js
louidorjp: Everything works fine. I just made a small update.
`const differnenceInYears = nextVisitDate.getFullYear() - encounterDate.getFullYear();`

mogoodrich:
Thanks @louidorjp  for the update and the approval... merging this in now!

PR #214, opened 2022-09-09
refactor to increase performance

dmdesimone:


configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
```diff
@@ -44,27 +47,47 @@ where pp.patient_program_id =
...
-         inner join temp_mch_pp pp on pp.patient_id = p.patient_id 
-;
+create index temp_J9_patients_pi on temp_J9_patients(patient_id);
+
+-- delete patients who are dead
+delete t from temp_J9_patients t
+inner join person p on p.person_id = t.patient_id
+where p.dead = 1;
+
+-- load all relevant encounters (for use later to narrow down observations)
```
mseaton, line 65: Shouldn't this table be called temp_j9_enc or something, since it includes both obgyn and delivery encounters?  Also, can we move this further down the query right before it is actually used, so it's easier to follow the progression of thought?

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
```diff
@@ -44,27 +47,47 @@ where pp.patient_program_id =
...
+create temporary table temp_obgyn_enc
+select e.encounter_id  from encounter e
+inner join temp_J9_patients t on t.patient_id = e.patient_id 
+where e.encounter_type in ( @obgynEncId, @delEncId)
+and e.voided = 0;
+
+-- uuid, age,  birthdate
+update temp_J9_patients t
+inner join person p on p.person_id = t.patient_id
+	set patient_uuid = p.uuid,
```
mseaton, line 76: patient_uuid -> t.patient_uuid

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
```diff
@@ -44,27 +47,47 @@ where pp.patient_program_id =
...
+		t.birthdate = p.birthdate;
+
+update temp_J9_patients t
+set t.age = TIMESTAMPDIFF(YEAR,t.birthdate, now());
```
mseaton, line 80: Do we not have a function for this - current_age_in_years?

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
```diff
@@ -44,27 +47,47 @@ where pp.patient_program_id =
...
+
+-- program state
+update temp_J9_patients t
+set t.J9_program = currentProgramState(t.patient_program_id,@MCH_treatment,'en');
 
 -- telephone number
 update temp_J9_patients t
```
mseaton, line 87: Let's be consistent and use the functions that are already well developed.  Don't we have a phone_number(_patient_id) function?

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
```diff
@@ -44,27 +47,47 @@ where pp.patient_program_id =
...
     left outer join person_attribute pa on pa.person_id = t.patient_id and pa.person_attribute_type_id = @tele
         and pa.voided = 0
-set t.telephone = pa.value
-;
+set t.telephone = pa.value;
 
 -- patient address
 update temp_J9_patients t
```
mseaton, line 93: What is this "current_name_address" ?  Do you encourage this use still?  Should we use functions instead for this?  (eg. person_address_city_village)?

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
```diff
@@ -73,47 +96,82 @@ set t.department = cna.department,
...
-set t.Dossier_identifier =  dosId(t.patient_id)
-;
+set t.Dossier_identifier =  dosId(t.patient_id);
 
 -- patient names
 update temp_J9_patients t
     left outer join current_name_address cna on cna.person_id = t.patient_id
 set t.given_name = cna.given_name,
-    t.family_name = cna.family_name
```
mseaton, line 91: Same here regarding use of current_name_address.

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
```diff
@@ -73,47 +96,82 @@ set t.department = cna.department,
...
+where o.voided = 0
+and o.concept_id = @mothersGroup; 
+
+create index temp_mg_obs_ci on temp_mg_obs(person_id, obs_datetime);
 
--- J9 group (latest observation of entered J9 group, since program enrollment)
 update temp_J9_patients t
-set t.J9_group =  (select value_text from obs where obs_id = latestObs(t.patient_id, @mothersGroup, j9_enrollment_date))
-;
+inner join obs o on o.obs_id =
```
mseaton, line 126: Why do you have to join back on the obs table here?  Is that a limitation of mysql and temp tables?  If so, why bother adding o.value_text and o.obs_datetime to temp_mg_obs, since you don't use them?

I wonder if it would be simpler (and not much less efficient) to just create a single temp_j9_obs table with _all_ of the observations in the temp_j9_enc table, and then use that in all of the subsequent queries.

dmdesimone: SO yeah, a few things: Mysql doesn't allow you to reopen a temp table in the same query like you can a non-temp table.  I tested out creating a copy of the temp_mg_obs table and using that, but it was just as quick to use the actual obs table, which isn't a surprise since it is just joining on obs_id and that is indexed.  

I tested out using just a single temp_j9_obs table (which how it seems to work best in other queries) but (at least on HUMTest), it ended up being faster to create individual temp tables for each of these.  

But yeah, I can remove value_text from this.  (and similarly from the other 2 times I do this)

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
dmdesimone: So yes, but in this case since I needed to look up the person table row anyway for the UUID (and I guess we could write a function for THAT also) , I figured it would be more efficient to update the birthdate there.  Once we have the birthdate, it's super quick to do any of these calculations, obviously.  

I definitely agree that we should standardize on functions as much as possible but my approach for THIS  query was just to reduce run time as much as possible since it's unique in that it runs ON the transactional database, unlike any of the other scripts.

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
dmdesimone: yeah, this is a view we wrote a while back:  https://github.com/PIH/openmrs-config-pihemr/blob/749d8416b957d93e5ce253659929f27b619499f7/configuration/pih/liquibase/sql/create_views.sql

It works and performs well, but it's only a few extra seconds (on HUMTest) to use the functions so I made the change.

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
dmdesimone: same as above

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
```diff
@@ -5,24 +5,29 @@ select concept_id into @edd from concept where uuid = '3cee56a6-26fe-102b-80cb-0
...
+    J9_group_max_datetime datetime,
     J9_group text,
     J9_program varchar(50),
     telephone varchar(50),
     birthdate date,
     age int(3),
+    edd_max_datetime datetime,
     estimated_delivery_date date,
+    add_max_datetime datetime,
     actual_delivery_date date,
```
mseaton, line 31: @dmdesimone I don't see where the new columns - J9_group_max_datetime, edd_max_datetime, and add_max_datetime are used.  We these added for something you didn't end up needing them for, and then never removed.  Can we remove them to eliminate any future confusion?

mseaton (approved):
One new comment about some table columns that don't appear to be used.  Otherwise looks good.

configuration/reports/reportdescriptors/dataexports/sql/j9CaseRegistration.sql
dmdesimone: yeah... exactly.  I went through an iteration of another approach which wasn't any faster.  I will remove these!

PR #221, opened 2022-10-11
Hai 733 le medecin du programme doit pouvoir voir le regime du patient

ljoedavessenn:
An issue happened when merging with the master. I corrected it.


lnball (approved):
Assume this was a duplicate id?

PR #224, opened 2022-11-09
UHM-6780 - Disabled submit and confirmation buttons should not change…

mseaton:
… to an active color on hover


mogoodrich (approved):
LGTM, thanks @mseaton !

PR #226, opened 2022-12-01
UHM-6887 - disable fingerprinting component

mseaton:


mogoodrich (approved):
Yep, thanks @mseaton !  There are some biometrics configurations somewhere as well, but might as well keep those in place in case we turn them back on again?

PR #231, opened 2023-02-01
Make dbevent setup happen based on a component.  Move initialization …

mseaton:
…into startup method.


mogoodrich (approved):
Fine with me... I assume you have updated the PIH Core PR to include this component?

PR #233, opened 2023-02-14
UHM-7016: Change user privileges for patient document (attachment)

mogoodrich:


mogoodrich (commented):
Note that there are some administrative roles, like "schedule manager" who had access to the patient dashboard, so I did go ahead and give them attachments access, but we could re-think.

lnball (commented):
LGTM.  Not sure about the schedule dashboard user.  Unlikely that they would need to see/add attachments.  Dave can weigh in tomorrow.

PR #234, opened 2023-02-16
HAI-709: Submit button disabled when phone number is incorrect

louidorjp:
Screenshot from 2023-02-16 10-41-48

This function "setUpNextAndSubmitButtons" does not work, in "patientRegistration-contact.xml". for this reason I added the function directly :> 
Screenshot from 2023-02-16 11-05-21XXXX

FYI @mogoodrich , @brandones , @lnball

### DIFF
--- a/configuration/pih/htmlforms/patientRegistration-contact.xml
+++ b/configuration/pih/htmlforms/patientRegistration-contact.xml
@@ -5,11 +5,6 @@
 
         <script type="text/javascript">
 
-            function setButtonsDisabled(val){
-                jq("#next").prop("disabled",val)
-                jq("#submit,.confirm").prop("disabled",val)
-            }
-            
             jq(function() {
 
                 var personAddressWithHierarchy = {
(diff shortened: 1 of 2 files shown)


configuration/pih/htmlforms/patientRegistration-contact.xml
```diff
@@ -5,6 +5,11 @@
 
         <script type="text/javascript">
 
+            function setButtonsDisabled(val){
+                jq("#next").prop("disabled",val)
```
mogoodrich, line 9: I don't think there's a next button on this form, so I'm not sure we need it?

mogoodrich (commented):
So @louidorjp  this generally looks good, but one risk is that I don't think it work correctly if there's something else that ends up needing to be validated on the same page?

For instance, say there were two phone numbers on this form, and the user typed them in wrong.  The submit button would be disabled, but as soon as they fixed one of the numbers, the submit button would be enabled again (I believe) and the user could submit with the other phone number incorrect.

mogoodrich (approved):
Works for me, thanks @louidorjp !

PR #238, opened 2023-04-13
INH initiation/completion date on HIV Patient Summary

louidorjp:
please review this PR.

I still have something to check but, not to block the process, I push the code to proceed with the migration @dmdesimone


mogoodrich:
Thanks @louidorjp !  I should be able to take a look tomorrow

configuration/pih/statusData/hivTb/inhStatus.sql
```diff
@@ -1,18 +1,22 @@
...
+select concept_from_mapping('PIH', '14534') into @inhStartDate;
+select concept_from_mapping('PIH', '14535') into @inhEndDate;
 
-select      ifnull(o.scheduled_date, o.date_activated),
-            ifnull(o.date_stopped, o.auto_expire_date)
-            into @startDate, @endDate
-from        orders o
-where       o.patient_id = @patientId
+
+select     value_datetime into @startDate
```
mogoodrich, line 7: defer to @dmdesimone  on the SQL

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -156,6 +156,243 @@
...
           });
+
+    });
+</script>
+</excludeIf>
+
+<excludeIf velocityTest="$fn.getObs($encounter, 'PIH:14535')">
+<script type="text/javascript">
+
+    jq(document).ready(function() {
```
mogoodrich, line 167: It might not be possible, but these three functions seem to share a lot of similar code, would be great if we could abtract that out into a utility function.

Also, would be helpful to add some comments saying what this is all doing, it's not immediately apparent to me.

mogoodrich (commented):
@louidorjp added a few comments, happy to look at this further when it's ready for review

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -283,7 +520,7 @@
 
                     <!-- Viral Load -->
                     <repeat>
-                        <template>
+                        <template>syph-checkbox
```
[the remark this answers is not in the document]
louidorjp: I think it's a typo.

configuration/pih/htmlforms/hiv/section-hiv-state.xml
louidorjp: I understand, that's why I mentioned in the comment that I have something to fix. I will do it.

mogoodrich: No worries, thanks @louidorjp !

configuration/pih/htmlforms/hiv/section-hiv-history.xml
```diff
@@ -1531,10 +1531,10 @@
                                     </p>
                                 </td>
                                 <td class="{drugname}-hx">
-                                    <obs conceptId="CIEL:1190" class="startDate"/>
+                                    <obs conceptId="CIEL:164852" class="startDate" allowFutureDates="true" allowPastDates="false"/>
```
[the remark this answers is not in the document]
dmdesimone: the concepts look good but yes, we should probably migrate old data for this.

configuration/pih/statusData/hivTb/inhStatus.sql
dmdesimone: looks good!

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -156,10 +156,248 @@
...
+           
+                if (lastRecordedTreatmenLine.date != null) {
+                  
+                   if(lastRecordedTreatmenLine.treatmenLine.display !==null){
+
+                      setValue('Isoniazide.value', 1505);
+                   
+                    jq('#treatmentLine input[type="radio"]').each(function(i,domEl){
+
+                         if(lastRecordedTreatmenLine.treatmenLine.display==="Primary" || lastRecordedTreatmenLine.treatmenLine.display==="Primaire"  &amp;&amp; i==0){
```
mogoodrich, line 376: If we have the uuid, it might be safer here to match on that instead of display name, in case the display name changes?  If we do keep the display, we should double-check this works in French, Kreyol and English... do we use "Primaire" in both the French and Kreyol locales?

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -156,10 +156,248 @@
...
+                      encounterDisplay: ""
+                    };
+                   
+
+                    if( (currentEncounterDate.getTime() >=  inhEndDateObsDate.getTime()) ||
+                        (currentEncounterDate.toDateString() === inhEndDateObsDate.toDateString())) {
+                            if (typeof inhEndDateObs.value !== "undefined" &amp;&amp; inhEndDateObs.value !== null &amp;&amp; (inhEndDateObs.value.length &gt; 0)) {
+                                tmpInhEndDateObs.endDate = inhEndDateObs.value;
+                                tmpInhEndDateObs.date = inhEndDateObsDate;
+                                tmpInhEndDateObs.encounterDisplay = inhEndDateObs.encounter.display;
```
mogoodrich, line 212: do we need variable, do we use it all?

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -156,10 +156,248 @@
...
+                    }
+                }
+            }
+
+           
+                if (lastRecordedTreatmenLine.date != null) {
+                  
+                   if(lastRecordedTreatmenLine.treatmenLine.display !==null){
+
+                      setValue('Isoniazide.value', 1505);
```
mogoodrich, line 372: Is 1505 the concept id?  If so, unfortunately, the concept ids aren't consistent across servers, so we'd likely need to do something a little more complex where we look up the concept id based on the uuid or mapping of the concept.

mogoodrich:
@louidorjp I added some thoughts, but defer you if it's ready to merge.  I agree that making this more generic can happen in another ticket.

Some of my comments are just suggestions... my main concern is that the line "setValue('Isoniazide.value', 1505)" likely won't work on all servers,  if 1505 is a concept id?

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -156,10 +156,248 @@
...
+<script type="text/javascript">
+
+    jq(document).ready(function() {
+        
+
+        const contextPath = window.location.href.split('/')[3];
+        const apiBaseUrl =  "/" + contextPath + "/ws/rest/v1";
+        const patientUuid = '<lookup expression="patient.uuid"/>';
+
+        let currentEncounterDate = new Date();
```
mseaton, line 175: Current encounter date shouldn't default to "today" if the form field isn't present, but rather it should default to the current encounter date of the existing encounter (if any) via velocity expression.

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -156,10 +156,248 @@
...
```
mseaton, line 246: Current encounter date shouldn't default to "today" if the form field isn't present, but rather it should default to the current encounter date of the existing encounter (if any) via velocity expression.

configuration/pih/htmlforms/hiv/section-hiv-state.xml
```diff
@@ -156,10 +156,248 @@
...
+<excludeIf velocityTest="$fn.getObs($encounter, 'PIH:13786')">
+<script type="text/javascript">
+
+    jq(document).ready(function() {
+        
+        const contextPath = window.location.href.split('/')[3];
+        const apiBaseUrl =  "/" + contextPath + "/ws/rest/v1";
+        const patientUuid = '<lookup expression="patient.uuid"/>';
+
+        let currentEncounterDate = new Date();
```
mseaton, line 318: Same comment.  There are some utility methods I started adding to help with things like this and to have this defined in one place rather than to have to re-invent/copy-paste everywhere.  See:  https://github.com/PIH/openmrs-config-pihemr/blob/master/configuration/pih/scripts/global/htmlFormFunctions.js

configuration/pih/htmlforms/hiv/section-hiv-state.xml
mseaton: I assume this can be done via a velocity expression to look up a concept by mapping or uuid.  I also started writing some javascript services that can be used / added to for utlities like this:  https://github.com/PIH/openmrs-config-pihemr/blob/master/configuration/pih/scripts/global/pihServices.js#L33

configuration/pih/htmlforms/hiv/section-hiv-state.xml
mseaton: Agree with @mogoodrich .  We can not use the display name here, as this is not something we can rely upon.  This should be changed to compare by uuid or mapping, using a utility method ideally that we can reuse elsewhere.

configuration/pih/htmlforms/hiv/section-hiv-history.xml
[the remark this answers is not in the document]
mseaton: +1.  Yes, we always need to keep in mind that if these forms have been used already, then we need to account for the existing data with the previous concepts.

mseaton (changes requested):

louidorjp:
After our meeting, @lnball , @dmdesimone and I, decided to keep the old concepts for the start  and end dates of the INH. I created a new query to make the start and end dates of the INH available on the patient dashboard.

@mogoodrich, the javascript work has been removed from this ticket and will be done in other tickets.

 The work assigned for this ticket was to add the INH start and end dates to the form, and then display them in the patient dashboard.

mogoodrich (commented):
Thanks @louidorjp !  Splitting it up works for me, I'll defer to Ellen and Dave on the HTML Form code and then SQL code...

configuration/pih/htmlforms/hiv/section-hiv-history.xml
```diff
@@ -1530,11 +1530,13 @@
                                              answerConceptIds="CIEL:159943, CIEL:159944" />
                                     </p>
                                 </td>
-                                <td class="{drugname}-hx">
-                                    <obs conceptId="CIEL:1190" class="startDate"/>
+                                 <td class="{drugname}-hx">
```
lnball, line 1533: @louidorjp Which concept do you want here?  Don't you want to keep the concept id so you don't need to migrate data?

configuration/pih/htmlforms/hiv/section-hiv-history.xml
```diff
@@ -1530,11 +1530,13 @@
...
+                                    <obs conceptId="CIEL:163526" class="startDate" allowFutureDates="true" allowPastDates="true"/>
                                 </td>
+                               
                                 <td class="{drugname}-hx">
-                                    <obs conceptId="CIEL:1191" class="endDate"/>
```
lnball, line 1537: @louidorjp Which concept do you want here?  Don't you want to keep the concept id so you don't need to migrate data?

lnball (commented):
See comments.  Nice work @louidorjp

configuration/pih/htmlforms/hiv/section-hiv-history.xml
louidorjp: Yes @lnball , I kept the existing concepts

PR #239, opened 2023-04-19
UHM-7163: Enable MFA on all PIH EMR instances

mogoodrich:


mogoodrich:
@mseaton  does this look correct?  I just moved the authentication config we had set up on HUM-CI to the top level configs, and removed totp (which I don't think we are using, and possibly doesn't work anyway?)

configuration/pih/pih-config-haiti.json
```diff
@@ -15,6 +15,17 @@
         "registrationapp.paperRecordIdentifierDefinition": "paperrecord.patientDataDefinition.paperRecordIdentifier"
     },
 
+    "authenticationConfig": {
+      "scheme": "2fa",
+      "schemes": {
+        "2fa": {
+          "config": {
+            "secondaryOptions": "secret"
```
mseaton, line 23: totp should also be added, unless there is a specific reason we are choosing not to include it @mogoodrich ?  why do you think it doesn't work?

configuration/pih/pih-config-mirebalais.json
```diff
@@ -15,6 +15,17 @@
     "registrationapp.paperRecordIdentifierDefinition": "paperrecord.patientDataDefinition.paperRecordIdentifier"
   },
 
+  "authenticationConfig": {
+    "scheme": "2fa",
+    "schemes": {
+      "2fa": {
+        "config": {
+          "secondaryOptions": "secret"
```
mseaton, line 23: Same here.

mseaton (changes requested):

configuration/pih/pih-config-haiti.json
mogoodrich: Perfect, I'll add it back it... no reason to think it doesn't work, I just had forgotten if we had built it or not.

mogoodrich:
Added totp back in, merging, thanks @mseaton

PR #241, opened 2023-05-02
UHM-7187 - Add dbevent component for ZL pih configurations

mseaton:


configuration/pih/pih-config-haiti-mentalhealth.json
```diff
@@ -21,6 +21,7 @@
     "chwApp",
     "clinicianDashboard",
     "dataExports",
+    "dbevent",
```
mogoodrich, line 24: we might want to hold enabling on the laptops to start (though I guess it wouldn't hurt?)

mogoodrich (approved):
Generally looks good to me, see one comment. I'm forgetting, does this immediately enable the event logging? If so, we should make sure to monitor for issues on the next deploy.

configuration/pih/pih-config-haiti-mentalhealth.json
mseaton: We might just want to get rid of the component altogether.  We can disable this via a system/runtime/global property on a server-by-server basis that doesn't require new config deployment, so probably best to limit that way if we want to hold off on the laptops.

mogoodrich: Probably a future ticket to remove the component, once we have it deployed everywhere?

PR #246, opened 2023-06-21
uhm-7239 tb prophylaxis  order_reason script

dmdesimone:


configuration/pih/migrations/tb_prophylaxis_order_reason.sql
```diff
@@ -0,0 +1,28 @@
...
+and o.order_action = 'NEW' 
+and o.voided = 0
+and not EXISTS 
+	(select 1 from orders o2
+	where o2.patient_id = o.patient_id
+	and o2.voided = 0
+	and o2.date_activated <= COALESCE(o.date_stopped,auto_expire_date)
+	and o2.date_stopped >= COALESCE(o.date_stopped,auto_expire_date)
+	and o2.order_action = 'NEW' 
+	and o2.concept_id IN
```
mseaton, line 20: Rather than limiting to these specific concepts (which won't cover everything, particularly any multi-ingredient combination drug), it might be better to limit to any order that has order_reason = 'TB Treatment' and which is not the same as the INH order that is being checked.

configuration/pih/migrations/tb_prophylaxis_order_reason.sql
```diff
@@ -0,0 +1,28 @@
...
+	and o2.order_action = 'NEW' 
+	and o2.concept_id IN 
+		(concept_from_mapping('PIH','RIFAMPICIN'),
+		concept_from_mapping('PIH','ETHAMBUTOL'),
+		concept_from_mapping('PIH','PYRAZINAMIDE')) 
+	); 
+
+update orders o
+inner join temp_tb_prophylaxis_orders t on t.order_id = o.order_id
+set o.order_reason = concept_from_mapping('PIH',14256);
```
mseaton, line 28: As a sanity check, may want to only update the order_reason if the existing order_reason = 'TB Treatment'?  Depends on the state of the existing data...

mseaton (commented):
Looks good, but a few comments for you to consider.

configuration/pih/migrations/tb_prophylaxis_order_reason.sql
dmdesimone: also I confirmed that for ANY of the regimens utilizing Isoniazide the order reasons are all over the place

dmdesimone: @mseaton ^^

PR #247, opened 2023-06-29
Update pregnant status after PTME Initial and followup which specifies pregnant

louidorjp:
Screenshot from 2023-06-29 13-04-57

Please review this PR


dmdesimone:
The requirements seem better to me (EDD in the future).  I don't know if we also need to check to see if there is a delivery form?  (since the EDD is not always the actual delivery date).  But if the clinicians agree, I am ok with it.

defer to @mogoodrich on the actual code review

louidorjp:
> The requirements seem better to me (EDD in the future). I don't know if we also need to check to see if there is a delivery form? (since the EDD is not always the actual delivery date). But if the clinicians agree, I am ok with it.
> 
> defer to @mogoodrich on the actual code review

There's a 2nd ticket that will verify the actual delivery date

configuration/pih/statusData/hiv/lastPregnancyStatus.sql
```diff
@@ -1,25 +1,13 @@
...
+where o.obs_id = latest_obs(@patientId, 'PIH', '5596')
+and o.voided = 0;
 
 select
-    @gender as gender,
-    if (@pregnantValueCoded = concept_from_mapping('CIEL', '1065'), true, false) as pregnant,
-    if (@pregnantValueCoded = concept_from_mapping('CIEL', '1065'), date(@edd), null) as dueDate
-;
+  @gender as gender,
+    if(@edd > now(),date(@edd),null) as estimatedDeliveryDate;
```
mogoodrich, line 13: This could become tricky to calculate, but, what if the baby is overdue? I think this assumes that once you reach and pass the delivery date, the patient is no longer pregnant, which isn't the case if the baby is delivered late?

mogoodrich:
> The requirements seem better to me (EDD in the future). I don't know if we also need to check to see if there is a delivery form? (since the EDD is not always the actual delivery date). But if the clinicians agree, I am ok with it.
> 
> defer to @mogoodrich on the actual code review

Oh, yeah, just seeing this comment now, but I had a similar question, see above.

configuration/pih/statusData/hiv/lastPregnancyStatus.sql
louidorjp: You're right, @mogoodrich . It's a tricky calculation. 
you can't base the calculation on the estimated date of delivery. The clinician or the patient may not know the last date of menstruation.

I think we need to consider other parameters for the calculation.

PR #248, opened 2023-07-13
Limiter au numero de recu numerique

Dutervil:
Please review this PR for me

### DIFF
--- a/configuration/pih/htmlforms/checkin.xml
+++ b/configuration/pih/htmlforms/checkin.xml
@@ -96,7 +96,7 @@
 
                 // hack to make the datepicker show when the form is opened
                 getField('encounterDate.value').datepicker('show');
-                checkNumberTest();
+               allowNumericalNumber()
 
             });
 
--- a/configuration/pih/htmlforms/checkin.xml
+++ b/configuration/pih/htmlforms/checkin.xml
@@ -96,6 +96,7 @@
 
                 // hack to make the datepicker show when the form is opened
                 getField('encounterDate.value').datepicker('show');
+                checkNumberTest();
 
             });
 
@@ -176,7 +177,9 @@
                                  answers="$paymentAmts" defaultValue="$paymentAmtDefault"/>
                         </p>
                 </fieldset>
-                <fieldset id="receipt-number-question">
+
+             
+                <fieldset id="receipt-number-question" class="required">
                     <legend>
                         <uimessage code="mirebalais.retrospectiveCheckin.receiptNumber.label"/>
                     </legend>
@@ -185,10 +188,14 @@
                     </h3>
 
                     <p class="left">
-                        <obs conceptId="PIH:Receipt number" id="receiptNumber" showUnits="false"
+                        <obs value="number" conceptId="PIH:Receipt number" id="receiptNumber" showUnits="false"
                              unitsCssClass="append-to-value" style="dropdown" />
+                        <br/>
+                        
                     </p>
+
                 </fieldset>
+
             </obsgroup>
 
             </section>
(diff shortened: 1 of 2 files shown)


configuration/pih/scripts/global/zl.js
```diff
@@ -222,4 +222,15 @@ function setupReturnVisitDateValidation(encounterDate, returnVisitDateLessThanEn
 
   jq(domEl).change(returnVisitDateValidation);
   beforeSubmit.push(returnVisitDateValidation);
+  
+
+
+}
+
+function allowNumericalNumber() {
```
mseaton, line 230: If we are adding a generic function to global/zl.js, we should make this take the id of the input as a parameter to the allowNumericalNumber function, and not hard-code #receiptNumber here.

configuration/pih/htmlforms/checkin.xml
```diff
@@ -185,10 +188,14 @@
                     </h3>
 
                     <p class="left">
-                        <obs conceptId="PIH:Receipt number" id="receiptNumber" showUnits="false"
+                        <obs value="number" conceptId="PIH:Receipt number" id="receiptNumber" showUnits="false"
```
mseaton, line 191: What does `value="number"` do?  This doesn't seem right, we don't want to set the value of the data to the text of "number".

configuration/pih/scripts/global/zl.js
mseaton: This isn't really _allow_ a numerical number, but really is intended to _restrict_ to a number, right?  Numbers are already allowed in the text input.  This function is intending to restrict the input to only a number.  So lets name the method accordingly.

configuration/pih/htmlforms/checkin.xml
mseaton: `style="dropdown"` doesn't really make sense here does it?  This is a text box.  Also `unitsCssClass` isn't really needed, since we have `showUnits="false"`, right?

Dutervil: You are Right

lnball: This concept is text (class = question; datatype = text).  Only datatype of numeric have units (ie. weight (kg) or blood volume (Liter).

image

As @mseaton suggested, style=dropdown is appropriate for datatype=coded where there are answers to a question (ie. Color of automobile = red, white, orange)

Dutervil: According to this Ticket https://pihemr.atlassian.net/browse/HAI-754
I had to allow only numeric value to the input
Before the change, the type was text as you say
this might be an error  in the ticket, I quess

mseaton: Your approach seems correct @Dutervil .  I don't think we need or want a new Concept, just want to restrict the input box to a "numeric", though this might have an adverse impact on editing existing data.  By using the existing Text Concept, you will get a text box that is not a number field, but your change to use JS to make this a number input seems reasonable.

mogoodrich (approved):
Thanks @Dutervil !  (and nice to meet you!)

Looks like @mseaton  has done a detailed review, but my quick review this looks good to me, thanks!

One side note... I think it's totally fine to do what you did and just look up by id, but there are also some "helper" functions provided by HTML Form Entry that allow you to fetch a field, these are documented below.  These can be particularly helpful for elements like dates/datepickers where it might not be clear what HTML element you need to fetch:

configuration/pih/htmlforms/checkin.xml
Dutervil: Thank for your approach @mseaton

Dutervil:
Thank your @mogoodrich 
I will have a look in the Doc

Dutervil:
I have to finish this ticket today!

mogoodrich:
I think this is fine to go ahead and merge @Dutervil when you are ready!

PR #252, opened 2023-08-07
HAI-855: add 'yes, but other reason' to Does patient require ART

louidorjp:
Please review this PR


mogoodrich (approved):
Generally defer to Ellen on form stuff, but looks good to me!

PR #261, opened 2023-09-07
UHM-7394 Added column to retire drugs

lnball:
This code is only in this branch.
* This code does NOT work.  This csv does NOT retire a drug.  
* Note the last 3 drugs with 1 in column 2 (Void/Retire).

By adding this to master, it will CREATE 3 new drugs (retired).  This worries me since we would need to test all the uses of drug orders to make sure that they behaved properly (show any drug ordered BUT don't show retired drugs when ordering)


mseaton (changes requested):
I commented on the ticket, but my opinion is that we should not do this just to clean up a few test servers.

lnball:
I am cleaning up the databases @mseaton 

As long as I added a column (void/retired), what about keeping that for a rainy day?  There's no data in the column.

If you think it's introducing a change for no reason, I get that too.

PR #272, opened 2023-11-09
fixing bug

Dutervil:


mogoodrich:
@Dutervil  for changes as small as this, it's okay not to have a ticket linked in the commit (though if this change is part of ticket, by all means include the ticket number), but going forward it would be good to have a little more informative commit message, like "fixing French translation" instead of just :"fixing bug".

PR #278, opened 2023-12-21
UHM-7624, add validation function conditionally if section is included

cioan:
This validation function was included even if the transmission section was not included for children. I have just moved the validation function within the conditional clause. Thanks @mogoodrich for reviewing this change!


mogoodrich (approved):
LGTM, thanks @cioan !

PR #287, opened 2024-04-04
HAI-940:Update INH display in clinical dashboard

louidorjp:
Please review this PR


mogoodrich (commented):
Thanks @louidorjp !  I defer to @dmdesimone  on this one... Dave, if you need me to look closer, please LMK...

dmdesimone (changes requested):
@louidorjp taking a closer look at this, I think you need to check for NULLs in the CASE statements.  Otherwise the comparisons will not work out if no data is entered in the orders or historical INH dates.

PR #291, opened 2024-05-13
CHW workflow fixed

louidorjp:
Please review this PR

### DIFF
--- a/configuration/pih/htmlforms/hiv/hiv-psychosocial.xml
+++ b/configuration/pih/htmlforms/hiv/hiv-psychosocial.xml
@@ -385,7 +385,7 @@
                             <label><uimessage code="general.fromDate" />:</label>
                             <span id="lastRecordedCHWEncounter" class="value"></span>
                             <br />
-                            <ifMode mode="EDIT" include="true">
+                            <ifMode mode="VIEW" include="false">
                                 <div>
                                     <label>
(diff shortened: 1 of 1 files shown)


configuration/pih/htmlforms/hiv/hiv-psychosocial.xml
```diff
@@ -385,7 +385,7 @@
                             <label><uimessage code="general.fromDate" />:</label>
                             <span id="lastRecordedCHWEncounter" class="value"></span>
                             <br />
-                            <ifMode mode="EDIT" include="true">
+                            <ifMode mode="VIEW" include="false">
```
mogoodrich, line 388: I'm unfamilar with this form, but this change will make the enclosed block included when both entering a new form and editing an existing form (previously it would only appear if editing) so if this is the correct behavior LGTM!  fyi @louidorjp @dmdesimone

mogoodrich (approved):

configuration/pih/htmlforms/hiv/hiv-psychosocial.xml
louidorjp: @mogoodrich  the reason I'm doing this is that CHW is also entered in hiv intake and followup. when entered in hiv intake and followup, for a first entry in the psychosocial form, the CWH name cannot be changed.

PR #293, opened 2024-07-11
Add Sante Mantal and Klinik MNT in Thomonde

andycottonpih:
Please review and let me know if this looks correct.


mogoodrich (approved):
Thanks @andycottonpih ! I didn't check the location tags in detail, would be good to make sure everything is matching up.

PR #297, opened 2024-08-02
UHM_8109 Changed component name to physicalRehab

lnball:
This goes with 2 other PRs.  See https://pihemr.atlassian.net/browse/UHM-8109


configuration/pih/pih-config-mirebalais-humci.json
```diff
@@ -16,6 +16,7 @@
     "dataExports",
     "dbevent",
     "dispensing",
+    "drugRehab",
```
mogoodrich, line 19: I think adding this will crash HUM-CI if it's merged in before the drug rehab component is added to PIH Core?

mogoodrich (commented):
See my comment

configuration/pih/pih-config-mirebalais-humci.json
lnball: @mogoodrich Good point.  It didn't crash on my local, but I've removed drugRehab.  Might add it back with the other code next week if we need a new form.

PR #303, opened 2024-10-28
Add Sal Timoun for Cange

andycottonpih:
Added location, locationtagmapping, and messageproperties for j9 (Sal Timon) at Cange.   I copied the locationtagmapping from La Colline, but removed the "TRUE," for Labs Component Location, which is not in the Cange mapping csv.

### DIFF
--- a/configuration/locations/locations.csv
+++ b/configuration/locations/locations.csv
@@ -78,6 +78,7 @@ Klinik MNT | HSN,47521fa1-ac7f-4fdb-b167-a86fe223a3d9,,The NCD clinic at HSN.,Sa
 Sante Mantal | Cange,61fcc62a-84c4-4d52-a3cd-25f58ff8c303,,Sante Mantal | Cange,Cange,,
 Sante Fanm | Cange,9aab3f94-eef5-4269-83a8-97cfe0679828,,'Women''s Outpatient clinic at Cange (Sante Fanm).',Cange,,
 Klinik MNT | Cange,3c609c2d-a9a2-4e59-932d-be53e6a39199,,The MNT clinic at Cange.,Cange,,
+Sal Timoun | Cange,397935c0-92e0-4d4c-be85-54f4447a8794,,Pediatrics at Cange,Cange,,
 Sante Mantal | Hinche,67b577e0-ff06-41a3-92f4-09dc27a5ae24,,Sante Mantal | Hinche,Hinche,,
 Sante Fanm | Hinche,602175ae-24d5-4711-bd28-511e4d702460,,'Women''s Outpatient clinic at Hinche (Sante Fanm).',Hinche,,
 Klinik MNT | Hinche,9705b785-0f00-41d1-a29b-3b357a307cdb,,The NCD clinic at Hinche.,Hinche,,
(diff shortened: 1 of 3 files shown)


configuration/locations/locations.csv
```diff
@@ -135,3 +135,4 @@ Swen Entansif | La Colline,199ed01b-d64f-4d19-aedf-ee37c7a83285,,ICU at La Colli
 Swen Entansif Neonatal | La Colline,85e4216d-7f78-42a1-bf5b-18967147db3e,,NICU at La Colline,La Colline,,
 Travay e Akouchman | La Colline,70048c2e-6d86-4f7c-89ff-d9a317758b27,,Labor and Delivery at La Colline,La Colline,,
 UMI | La Colline,d50694a1-d4b3-4da2-bf84-1c82b9a0f758,,Inite maladi enfeksyon (UMI) at La Colline,La Colline,,
+Sal Timoun | Cange,397935c0-92e0-4d4c-be85-54f4447a8794,,Pediatrics at Cange,Cange,,
```
mogoodrich, line 138: For clarity, instead of adding at the end of the file, can we add it up with the other Cange locations?  It shouldn't affect loading.

mogoodrich: And not to state the obvious but I assume you generate a new uuid using a uuid generator? (In the past I've sometimes copied an old line and forgotten to update the uuid)

configuration/locationtagmaps/locationtagmaps-site-cange.csv
```diff
@@ -3,3 +3,4 @@ Cange,,,,,,,,,,,,,,,,,,,TRUE,,,,,,,TRUE,,,,,,,,TRUE,
 Sante Mantal | Cange,,,,,,TRUE,,TRUE,,,,,,,,,TRUE,,,TRUE,,,,,,,TRUE,TRUE,,,,,,,TRUE
 Sante Fanm | Cange,,,,,,TRUE,,TRUE,,,,,TRUE,,,TRUE,TRUE,TRUE,,,,,TRUE,TRUE,,,TRUE,TRUE,,,,,,,TRUE
 Klinik MNT | Cange,,,,,,TRUE,,TRUE,,TRUE,,,,,,TRUE,TRUE,,,,TRUE,,,,TRUE,,TRUE,TRUE,,,,,,,TRUE
+Sal Timoun | Cange,TRUE,TRUE,,,TRUE,,,TRUE,,,,,,,TRUE,TRUE,TRUE,TRUE,,TRUE,,,TRUE,TRUE,TRUE,,TRUE,,,,,TRUE,,,TRUE
```
mogoodrich, line 6: Do we have inpatient functionality at Cange?  If not, probably makes sense to remove the admission and transfer tags (though if the ADT component is turned off at Cange it shouldn't make much of a difference).  Also, I would think that Sal Timoun would be a Check-In location, but it does look like you just copied this from Thomonde Sal Timoun, so maybe not.

mogoodrich (changes requested):
Thanks @andycottonpih !

One request about moving the new location earlier in the locations.csv file.

I made some comments about the location tags, but we don't necessarily need to change those if we don't have clear answers about how the locations will be used.

configuration/locationtagmaps/locationtagmaps-site-cange.csv
andycottonpih: I actually copied it from La Colline, since LC's locations are a superset of whats at Cange, and removed the one column that doesn't exist here, but otherwise did not make any changes.

configuration/locations/locations.csv
andycottonpih: I have updated the order as requested, and yes, this is a new uuid.

configuration/locationtagmaps/locationtagmaps-site-cange.csv
mogoodrich: Works for me

PR #308, opened 2024-11-21
UHM-8236: Review Program Workflow Iniz Deployment

mogoodrich:
see: https://pihemr.atlassian.net/browse/UHM-8236

note that this *must* be merged in in conjunction with the corresponding change to config-pihemr


mseaton (approved):

constants.yml
```diff
@@ -44,17 +44,3 @@ concept:
     uuid: "d087f783-de30-4aea-8a4d-50ec72ed656a"
   pedsIndividual:
     uuid: "59f75bbf-4a6f-42b9-92d4-9259cba548d2"
-
```
lnball, line 47: This was added by me.  Can't we use this more readable format?  You use these variables in the programWorkflowStates .csv file

@mogoodrich

mogoodrich: @lnball  I moved all the constants into the PIH EMR constants file to have them in a central place.  We can discuss whether that was the right approach or not, but I was thinking putting all the uuids declarations in a central place so those are shared across countries, if used, but then define the programs/workflows/states in the separate country configs.

PR #309, opened 2024-12-04
adding additianal field to vital sign form

Dutervil:
adding additianal field to vital sign form
@lnball


lnball (commented):
This is great Wadson.  Thanks for moving the vitals to the zl configuration.

From our conversation with Jeune on Monday, he suggested that some of the questions not appear everywhere (ie. Oxygen flow and Glasgow only for inpatient, ICU, etc; BDCF only for pregnant women).  We should discuss how to handle this.

In addition, you could have all the Pain scores on 1 tab instead of having tabs for each pain score.

lnball (approved):
Looks good

PR #318, opened 2025-02-25
remove Foetal Hearth Rate field from 'Sal Fanm'

Dutervil:
remove Foetal Hearth Rate field from  '' Sal apre Akouchman  
it should only be in 'Sante Fanm"


lnball (approved):
Looks good @Dutervil

PR #319, opened 2025-03-11
SL-859: Export an all patients list nightly and save offline

mogoodrich:


mseaton (commented):
Same comment here about the filenameBase.  Otherwise LG.

PR #320, opened 2025-03-13
HAI-1034 - Allow submission of consult note with only non-coded diagn…

mseaton:
…oses


lnball (approved):
This looks right...

mogoodrich (approved):
LGTM, thanks @mseaton

PR #337, opened 2025-04-28
section 2) EXAMENS DE LABORATOIRE msppReport

Dutervil:
I was using Lab order form for this section


mogoodrich (commented):
Thanks @Dutervil !  I defer to @dmdesimone  on the code, but I noticed a random lock file that it looks like shouldn't be committed.

louidorjp (approved):

dmdesimone (approved):
This looks good @Dutervil .  I didn't check all the concepts but the general approach looks good.  You can merge it and it will be tested!

PR #338, opened 2025-05-07
HAI-1033: MSPP : Section  Urgences

louidorjp:
Please review this PR


configuration/reports/reportdescriptors/dataexports/sql/msppReport.sql
```diff
@@ -565,6 +567,207 @@ SELECT  (@malaria_test_microscopique_vivax +
          ) INTO @malaria_test_microscopique_total ;
          
 
+-- PHYSICAL VIOLENCE
+SELECT 
+SUM(IF(concept_name(disp.value_coded, 'en') = 'Death' AND p.gender ='F',1,0)),
```
dmdesimone, line 572: it would be better to use the concepts for these dispositions, in case the names ever change.

E.g. use disp.value_coded = concept_from_mapping('PIH',DEATH') instead of  concept_name(disp.value_coded, 'en') = 'Death'

Or actually it would be more efficient to set @death = concept_from_mapping('PIH',DEATH');  and then use "if(disp.value_coded =@death..." or something.  

The mappings are more stable than the names are and should be used.

configuration/reports/reportdescriptors/dataexports/sql/msppReport.sql
```diff
@@ -609,5 +812,94 @@ SELECT SUM(child_under_1_n) "CHILD_UNDER_1_N",SUM(child_under_1_s) "CHILD_UNDER_
             @malaria_test_microscopique_oval'OVAL',
             @malaria_test_microscopique_malariae 'MALARIAE',
             @malaria_test_microscopique_falciparum 'FALCIPARUM',
-            @malaria_test_microscopique_total 'MALARIAN_TEST_MICRO'
+            @malaria_test_microscopique_total 'MALARIAN_TEST_MICRO',
+
```
dmdesimone, line 816: my SQL client did not like these blank lines in the middle of the query.  I needed to remove them to get this to run.  Not sure if it will cause a problem in OpenMRS.

dmdesimone (changes requested):
@louidorjp made just a few comments inline this

dmdesimone (approved):
Looks good!

PR #339, opened 2025-05-08
UHM-8535 Add config for Boucan Carre MH

andycottonpih:
Boucan Carre MH config based on the json added for Belladere MH.


mogoodrich (approved):
LGTM!

PR #344, opened 2025-05-22
HAI-1041: I have updated 

Dutervil:
HAI-1041: I have updated the delivery section report by using the the right form which is delivery form


dmdesimone (approved):
Looks good!

PR #347, opened 2025-05-29
Migrate Petite Riviere from MH laptop

andycottonpih:
These are copied from the Verrettes update, only the names have been changed.  It didn't look like anything needed adjustment, but please let me know if additional changes are needed to make the site unique.


configuration/pih/pih-config-haiti-petitereviere.json
```diff
@@ -0,0 +1,39 @@
+{
+  "welcomeMessage": "Bienvenue dans l'EMR de Petite Reviere",
+  "site": "PETITEREVIERE",
```
mogoodrich, line 3: typo

mogoodrich (changes requested):
Thanks @andycottonpih !  A couple comments... mainly, I think it's spelled "Riviere" (though I think we've misspelled it elsewhere previously).  

If for some reason we confirm that it should be spelled with an "e" then we will have to fix the existing places in the config-ZL where it is spelled with an i.

Also note that you will need to change the two files names with "Petite Reviere" in the name as well. 

(Specifically, just as an FYI, the "site" value set in pih-config must be a case-insensitive to the last part of the "{domainobject}-site-{site_name}" files, because that is how it knows which files to load.)

PR #373, opened 2025-07-21
UHM-8692 Belladere_mh Location Tags

andycottonpih:
Now that Belladere has multiple login locations Belladere-mh needs a separate location tag file.


lnball:
The code looks fine.  Seems like a similar approach to Cerca la source.  But I don't understand what you need -- is this for a new Belladere server?   

I'm sure it's fine, but please explain and I will approve.

andycottonpih:
Previously both the Belladere systems, NCD and Mental Health, used the same locationtag file with only one login location after being migrated to the colo.  Now the system that was NCD has many additional login locations that are not desired on the Mental Health system, so it needs a new file with just the one top level login location.  This ticket is just to make sure that is created before it becomes urgent.  Thanks!

PR #386, opened 2025-10-03
HAI-1067: create a separate sql file for estimated delivery date

Dutervil:
HAI-1067: create a separate sql file for estimated delivery date


dmdesimone (approved):
This looks good.  You could have kept it in the same .sql file but it works like this also!

PR #387, opened 2025-10-06
HAI-1067: Optimized query by removing subquery and improving performance using …

Dutervil:
Optimized query by removing subquery and improving performance using  date range and index. @dmdesimone


mogoodrich:
Thanks @Dutervil !

@dmdesimone  if this looks good to you, we can merge this in after the release and see if this helps with the loading issues.

mogoodrich (approved):
Thanks @Dutervil !  Merging this in (and re-enabled) so we can test the performance.

PR #431, opened 2026-04-20
UHM-9253 HIV and PMTCT followup are broken because of the section-hiv…

lnball:
…-state code

This section still has an error associated with Isoniazid.  When you click the checkbox, "save" and "next" buttons do not work.


mogoodrich (approved):
LGTM, thanks @lnball !

PR #444, opened 2026-07-13
HAI-1177: C section -Births by Age Group of Mothers

louidorjp:
Please review this PR


dmdesimone (approved):
This looks good!

Just a quick note - you might find that it runs faster if the concept_id as defined by "concept_from_mapping()" are stored as variables outside of the statement.  You would think that the SQL engine would cache the results and not have to rerun the function but I've found that is not always the case. 

No need to look into it unless it is performing poorly.

louidorjp:
> This looks good!
> 
> Just a quick note - you might find that it runs faster if the concept_id as defined by "concept_from_mapping()" are stored as variables outside of the statement. You would think that the SQL engine would cache the results and not have to rerun the function but I've found that is not always the case.
> 
> No need to look into it unless it is performing poorly.

Good point. I’ll adopt this approach.

PR #446, opened 2026-07-16
UHM-9334 Change location of frontend config

chibongho:
See: https://github.com/PIH/openmrs-frontend-pihemr/pull/13


mseaton (approved):
Fine but why get rid of the esm-login-app config?

chibongho:
> Fine but why get rid of the esm-login-app config?

Yeah, I can leave it in. We're not using it now, but will be if we ever migrate the login to O3.

PR #448, opened 2026-07-23
UHM-9361: display the Country field on the confirmation screen

cioan:
This just enables displaying the Country field on the Confirm screen.


mogoodrich (commented):
I assume there was a request to disable showing "Haiti" in some cases and that's why we set it up this way, but we can see if that's still what people want.

PR #454, opened 2026-08-12
TASKS-534 - Convert openmrs-config-zl to zl-emr with content and distro submodules

mseaton:


lnball (approved):
Leaving this to you @mseaton .  I'm not planning to review 246 file moves.

PR #458, opened 2026-08-13
HAI-1223: Populate Health Center on all servers and use in patient li…

mogoodrich:
…st reports

https://pihemr.atlassian.net/browse/HAI-1223

### DIFF
--- a/content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
+++ b/content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+
+    <changeSet id="20260810-backfill-health-center-attribute" author="mgoodrich">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">
+                select count(*) from global_property where property = 'pihcore.site' and (property_value = 'CROSS_SITE' or property_value='CENTRAL');
+            </sqlCheck>
+        </preConditions>
+        <comment>
+            HAI-1223: Backfill "Health Center" person attribute for existing patients missing it,
+            using the single most common visit location on this server. Skipped entirely on the
+            HIV/cross-site server (pihcore.site = CROSS_SITE).
+        </comment>
+        <sql>
+            select user_id into @adminUser from users where system_id = 'admin';
+            select person_attribute_type_id into @healthCenterType from person_attribute_type where uuid = '8d87236c-c2cc-11de-8d13-0010c6dffd0f';
+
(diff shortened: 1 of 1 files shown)


content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
```diff
@@ -0,0 +1,47 @@
...
+                select count(*) from global_property where property = 'pihcore.site' and (property_value = 'CROSS_SITE' or property_value='CENTRAL');
+            </sqlCheck>
+        </preConditions>
+        <comment>
+            HAI-1223: Backfill "Health Center" person attribute for existing patients missing it,
+            using the single most common visit location on this server. Skipped entirely on the
+            HIV/cross-site server (pihcore.site = CROSS_SITE).
+        </comment>
+        <sql>
+            select user_id into @adminUser from users where system_id = 'admin';
```
mseaton: Is this more reliable than using user_id = 1?

content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
```diff
@@ -0,0 +1,47 @@
...
+            create temporary table temp_most_common_visit_location
+            select location_id
+            from visit
+            where voided = 0
+            group by location_id
+            order by count(*) desc, location_id asc
+            limit 1;
+
+            insert into person_attribute (person_id, value, person_attribute_type_id, creator, date_created, voided, uuid)
+            select p.person_id, tmcvl.location_id, @healthCenterType, @adminUser, now(), 0, uuid()
```
mseaton: Can we evaluate now() once and use it as a variable, so these all end up with the exact same timestamp?

content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
```diff
@@ -0,0 +1,47 @@
...
+            from person p
+            inner join patient pt on pt.patient_id = p.person_id
+            cross join temp_most_common_visit_location tmcvl
```
mseaton: What the heck is a cross join and why are you choosing to use it here?  Why do we even need a temporary table for this location_id - can't we just read it into a variable and then use that variable everywhere?  I don't love this table + cross join thing for a single value that will be the same for everyone.

content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
```diff
@@ -0,0 +1,47 @@
...
+            where p.voided = 0
+            and not exists (
+                select 1 from person_attribute pa
+                where pa.person_id = p.person_id
+                and pa.person_attribute_type_id = @healthCenterType
+                and pa.voided = 0
```
mseaton, line 40: If you've tested this `not exists select 1` thing, and it works, fine.

content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
mogoodrich: Fair, will change

content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
mogoodrich: Fair, I can change

content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
mogoodrich: It's basically the same as doing "from person p, temp_most_common_location tmcvl"... as you probably guessed, a Claude thing, but seemed safe and not risky to have a join against a single value, but I guess using a variable instead would provide another layer of safety.

content/configuration/backend_configuration/pih/liquibase/liquibase-haiti.xml
mogoodrich: Yeah, I ran twice and confirmed it didn't double-add attributes.

mogoodrich (commented):
Fair comments from @mseaton ... I changed and tested locally, merging this in now...

PR #463, opened 2026-08-24
HAI-1230 Updated liveCheckin from checkin.  This form might be retire…

lnball:
…d so not so important


lnball:
The correct code changes are in the HAI-1230A branch

PR #468, opened 2026-08-31
UHM-9493: fix 'Screen for syphilis' date field hidden when checked

chibongho:
openmrs/openmrs-module-htmlformentry/pull/349 made a fix to `<obs style="checkbox" dateLabel="blah">` element to only show the date label and date field for the obs when the checkbox is checked. This conflicts with the custom `setUpObsWithObsDateTime` function here, which has its own logic to hide the date label and date field. This PR takes that stuff out, and only keeps the logic to limit the date field's `maxDate` to current date.

I confirmed that the `setUpObsWithObsDateTime` function is only in this repo, and is only used for the `#syph-checkbox` obs element.

Testing done:
In patient's visit dash, opened PMTCT-followup.xml form (in ENTER mode), which immediately saves. Reopen form in EDIT mode to verify that, when the checkbox is not checked,  the date label and field are not shown:

image

After checking the box, entering a date and save, reopen and form and confirm that the date label and field are shown and populated with the entered value:
image

### DIFF
--- a/content/configuration/backend_configuration/pih/scripts/global/zl.js
+++ b/content/configuration/backend_configuration/pih/scripts/global/zl.js
@@ -171,19 +171,12 @@ function phoneNumberPattern() {
  * @param widgetId
  */
 function setUpObsWithObsDateTime(widgetId) {
+  // The show/hide/clear of the obs date field for a checkbox obs is now handled natively by the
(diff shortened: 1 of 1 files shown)


mseaton (approved):
Taking your word for this

